PR #1814, opened 2026-06-07
ci(kani): pin kani-version + key toolchain cache on it (audit fix #1, clean rebase)

### LINKED ISSUE #1: Enforce monotone egress: detect and fail-closed on netns iptables drift
## Problem
We claim a monotone/immutable security envelope, but host netns iptables rules can be modified at runtime without detection. That means egress policy can silently relax after pod creation, violating the monotone posture.

## Why it matters
- Breaks the "static envelope" guarantee.
- Makes network policy a best-effort config instead of enforcement.

## Proposed approach
- On Firecracker pod start, snapshot the expected iptables rules for the pod netns.
- Periodically verify rules (or on each tool request) and fail-closed on drift (kill pod + revoke proxy).
- Optionally emit an audit log entry when drift is detected.

## Definition of Done
- Drift detection is enabled by default for Firecracker netns pods.
- Any deviation from the expected allow/deny rules triggers pod termination.
- An audit entry is recorded with pod id + diff summary.
- Add an integration test that simulates drift and verifies termination.

## References
- netns policy application: crates/nucleus-node/src/net.rs
- pod lifecycle: crates/nucleus-node/src/main.rs
- monotone posture docs: docs/architecture/overview.md, docs/assurance/hardening-checklist.md

brandon-coproduct:
**CI-efficiency audit fix #1 (re-cut clean off current main).** Supersedes #1813, which was branched before the MSRV-floor fix (#1811) merged and couldn't be updated in place (branch protection blocks force-push + merge-commits).

The 4 Kani jobs cached `~/.kani` keyed on `hashFiles('kani-nightly.yml')` → any edit re-downloaded the ~1 GB toolchain (~5–10 min/PR). Fix:
- Key the cache on the pinned Kani version (`kani-toolchain-<os>-kani-0.67.0`) instead of the file hash.
- Pin `kani-version: "0.67.0"` on all 4 steps (was unpinned `latest` → silently changes what the proof gate verifies; pinning makes proofs reproducible). Bump version + key suffix together on upgrade.

Off current main (MSRV 1.93 floor), so the Kani job now passes too. CI-config only.

### DIFF
--- a/.github/workflows/kani-nightly.yml
+++ b/.github/workflows/kani-nightly.yml
@@ -40,10 +40,18 @@ jobs:
         uses: actions/cache@668228422ae6a00e4ad889ee87cd7109ec5666a7 # v5.0.4
         with:
           path: ~/.kani
-          key: kani-toolchain-${{ runner.os }}-${{ hashFiles('.github/workflows/kani-nightly.yml') }}
+          # Key on the PINNED Kani version, NOT the workflow file: hashing the
+          # whole file re-downloaded the ~1 GB toolchain on every unrelated edit.
+          # Bump this suffix together with the `kani-version` input below when
+          # upgrading Kani.
+          key: kani-toolchain-${{ runner.os }}-kani-0.67.0
       - name: Run Kani (fast harnesses)
         uses: model-checking/kani-github-action@f838096619a707b0f6b2118cf435eaccfa33e51f # v1.1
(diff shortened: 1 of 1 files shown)


brandon-coproduct:
Duplicate — #1813 (identical change) already merged to main before this was cut. Closing.

PR #1826, opened 2026-06-10
feat(receipt): nucleus-receipt — the colimit envelope moves upstream

brandon-coproduct:
## What

New crate `crates/nucleus-receipt`: the **Session / Projection / Receipt** colimit envelope, ported wire-compatible from nucleus-platform's `nucleus-substrate-core`. Platform will re-export these types in a follow-up PR (its crate keeps the mechanism/discovery modules — those are mechanisms, not laws).

## Why

The 2025–26 agent-security SOTA converged on structural defenses (CaMeL, FIDES, Progent), and the open slot is a **verified composition**: label-lattice IFC + machine-checked confinement + signed action receipts on a verified substrate. The receipt envelope is the law layer of that composition and belongs in the open-source repo, where the verifier SDK trio (Rust / WASM-JS / Python) can verify every receipt kind — flow, constitutional, economic, identity — through one code path.

This is step 1 of the priority order: receipt colimit upstream → attenuation algebra → delegation category (RFC 8693 mapping) → recompute-reduction theorems per projection.

## Wire compatibility

- `canonical_signing_bytes` = sorted-key JSON of `{projections, session, version}`, identical to platform's; pinned by a **golden byte-equality test** so a transitive `serde_json/preserve_order` flip fails CI loudly instead of silently splitting signers from verifiers.
- `RECEIPT_VERSION = 1`, adjacently-tagged `{kind, body}` projections, BLAKE3 root hash, Ed25519 signature — all unchanged.
- Same `ed25519-dalek =3.0.0-pre.7` pin as the rest of the workspace.

## Tests

7 unit tests (round-trip, 3 tamper cases, wire-format stability, kind-string stability, golden canonical bytes) + 1 doctest. Pre-commit clippy/deny/audit green locally; full-workspace tests deferred to CI (local pre-push build exceeded disk).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6955,6 +6955,7 @@ dependencies = [
  "hex",
  "serde",
  "serde_json",
+ "serde_json_canonicalizer",
  "thiserror 2.0.18",
 ]
 
@@ -9834,6 +9835,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "6518fc26bced4d53678a22d6e423e9d8716377def84545fe328236e3af070e7f"
 
 [[package]]
+name = "ryu-js"
+version = "1.0.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "dd29631678d6fb0903b69223673e122c32e9ae559d0960a38d574695ebc0ea15"
+
+[[package]]
 name = "rzup"
 version = "0.5.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


brandon-coproduct:
CI caught a real one: the golden canonical-bytes test passed standalone but failed in the full-workspace build, because **cedar-policy-core unifies `serde_json/preserve_order` ON** there — i.e. plain `serde_json::to_vec` key order depends on the build graph, which is a latent signer/verifier split for any cross-binary verification.

Fixed in 21af9dfa by switching `canonical_signing_bytes` to **RFC 8785 (JCS)** via `serde_json_canonicalizer`: sorted keys at every depth, canonical numbers, independent of feature unification and of `Value` insertion order (regression test added). Verified locally both standalone and co-built with `nucleus-trust-registry` (the cedar pull). PR-2 will align platform's substrate-core to the same function via re-export, closing the same latent split there.

PR #1827, opened 2026-06-10
ci: feature-matrix gate — cargo-hack each-feature + MSRV floor check

brandon-coproduct:
## What

Two new CI gates in one workflow (`feature-matrix.yml`):

1. **`cargo hack check --each-feature --no-dev-deps`** over the eight security-load-bearing, feature-bearing crates (portcullis, portcullis-core, nucleus-lineage, nucleus-tool-proxy, nucleus-ifc, nucleus-envelope, nucleus-identity, nucleus-creditworthiness). Tests default / no-default / each feature individually — the consumer's-eye standalone build.
2. **MSRV floor**: `cargo check --workspace --all-features --locked` on 1.93 — the Kani-pinned `rust-version` was previously declared but unenforced.

## Why now

Feature unification is this codebase's recurring verified-claims hazard, not a style nit: cedar flips `serde_json/preserve_order` in full-workspace builds (caught **live** by nucleus-receipt's golden canonical-bytes test in #1826 — same source, different signing bytes standalone vs workspace), and the dalek de-vendor work hit pkcs8/alloc narrow-build gaps the same way. This makes "standalone build works" a tested property instead of a hope.

Part of the CI SOTA gap-closure series (next: zizmor gates, semver-checks, nextest).

### DIFF
--- a/.github/workflows/feature-matrix.yml
+++ b/.github/workflows/feature-matrix.yml
@@ -1,0 +1,62 @@
+name: Feature Matrix
+
+# Feature unification is a verified-claims hazard, not a style issue:
+# a crate that only ever builds inside the full workspace inherits
+# whatever features its neighbors unify ON (cedar flips
+# serde_json/preserve_order, jsonwebtoken pulls dalek pkcs8/alloc, …),
+# so "works in CI" says nothing about the standalone build a downstream
+# consumer gets. nucleus-receipt's canonical-bytes split (PR #1826) is
+# the canonical incident. This gate builds each security-load-bearing
+# crate per-feature so narrow builds are a tested property.
+
+on:
+  push:
+    branches: [main]
+  pull_request:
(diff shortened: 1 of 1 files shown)


brandon-coproduct:
First run of the gate caught a real one: **`portcullis --no-default-features` has never compiled** — `kernel.rs`/`certificate.rs` unconditionally import `crate::token`/`crate::delegation` (the kernel structurally needs `SessionProvenance`/`MeetJustification`), but both module declarations were `#[cfg(feature = "serde")]`. The modules were already internally `cfg_attr`-disciplined, so the fix is just ungating the declarations (150be097). Also excluded `nucleus-verifier-service` from the MSRV job — it `include_str!`s the CI-built wasm-pack artifact, so the floor was failing on artifact presence rather than source. All 8 crates × each-feature green locally.

PR #1830, opened 2026-06-10
feat(portcullis-core): the attenuation algebra — 'authority only tightens', named once

brandon-coproduct:
## What

New `portcullis-core::attenuation` module — the survey's abstraction #1, first slice (Rust):

- **`Attenuation<L: Lattice>`** — a monotone deflationary endomap (`f(x) ≤ x`, `x ≤ y ⇒ f(x) ≤ f(y)`): the single object that capability meet, delegation narrowing, budget caveats, constitutional monotonicity, and (dually) IFC taint all instantiate.
- **`MeetCap`** (canonical instance), **`Compose`** (laws closed under composition), **`chain_effective_authority`** — end-of-chain authority = meet of all caps, **independent of application order** (the RFC 8693 act-chain / SPIFFE multi-hop keystone: any verifier folds the chain in any order and agrees).
- **`verify_attenuation_laws`** sample-checker in the existing `verify_lattice_laws` style.

## The delegation instance, honestly scoped

`DelegationScope`'s order is glob-coverage but `intersect` is literal — an under-approximate meet, not a GLB (`["src/**"] ∩ ["src/lib.rs"] = ∅`). So the lattice claim is restricted to the **literal fragment** via a `LiteralDelegation` newtype that rejects glob scopes; on that fragment a parity test pins `narrow(p,c) = Some(·) ⟺ c ≤ p` with the narrowed value equal to the lattice meet. On the glob fragment, narrowing stays *deflationary* (sound, never widens) — meet-optimality is explicitly not claimed. Found along the way: lattice equality must be scope **set**-equality (deriving `PartialEq` breaks join commutativity via `Vec` ordering).

## Next

PR-3b adds `lean/PortcullisCore/Attenuation.lean` (composition closure + chain order-independence, machine-checked), then re-derives ck-policy monotonicity as an instance.

6 new tests; `cargo hack --each-feature` green on the crate; clippy clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1831, opened 2026-06-10
feat(lean): AttenuationProofs — the attenuation algebra, machine-checked (PR-3b)

brandon-coproduct:
## What

`lean/PortcullisCore/AttenuationProofs.lean` — the Lean side of #1830's `attenuation.rs`, **generic over any meet-semilattice** (where `DelegationCategoryProofs.lean` proves the binary laws for one concrete 3-field model):

- `Attenuation` structure with deflationary + monotone **proof fields** (the olog `AuthFibration.lean` pattern) and `comp` closure
- `meetCap_comp_collapse` / `meetCap_comp_comm` — chains of caps collapse to a single cap, in either order
- `chainAuthority_le_start` — a chain never grants authority
- **`chainAuthority_perm_invariant`** — the keystone: end-of-chain authority is invariant under permutation of the caps; any verifier may fold an RFC 8693 act-chain / SPIFFE delegation chain in any order and agree
- `applyChain_eq_chainAuthority` — holds by `rfl`: sequentially applying the caps as attenuations IS the single meet-fold (the recompute-reduction shape: verify a chain with one fold, not a replay)

Kernel-checked, **no `sorry`**; registered as `lean_lib AttenuationProofs`; built locally against the pinned mathlib v4.30.0-rc2 cache (339 jobs, new lib 10s). Olog prior-art review informed the proof-field pattern; the list-level perm-invariance theorem is new to both repos.

Follow-up: re-derive `ck-policy::check_monotonicity` as an instance, then the delegation-category PR (RFC 8693 mapping).

### DIFF
--- a/crates/portcullis-core/lean/AttenuationProofs.lean
+++ b/crates/portcullis-core/lean/AttenuationProofs.lean
@@ -1,0 +1,116 @@
+/-
+# The attenuation algebra — generic, machine-checked
+
+Lean counterpart of `src/attenuation.rs` (`Attenuation<L: Lattice>`,
+`MeetCap`, `Compose`, `chain_effective_authority`). Where
+`DelegationCategoryProofs.lean` proves the *binary* composition laws
+for one concrete constraint model (three min-composed fields), this
+file proves the laws **generically over any meet-semilattice** and
+adds the list-level keystone the Rust tests sample-check:
+
+1. `Attenuation.comp` — deflationary + monotone endomaps are closed
+   under composition (the structure-with-proof-fields pattern from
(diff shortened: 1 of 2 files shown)


brandon-coproduct:
The `cargo hack --each-feature` failure here is **pre-existing main breakage**, not this PR: the merge queue took #1827 with only its workflow commit (feature-matrix isn't a required check, so the armed auto-merge fired before the portcullis fix commit landed — the late push silently recreated the deleted branch on the closed PR). #1832 re-lands the fix plus the `-D warnings` item-level gates and will clear this. This PR's own change is Lean-only.

PR #1965, opened 2026-07-06
Reduce behavioural-consistency gaps in crate nucleus-cli

brandon-coproduct:

Reduce behavioural-consistency gaps in crate nucleus-cli ONLY. Touch ONLY files under crates/nucleus-cli/. CRITICAL — the `persona consistency` binary is permission-gated and CANNOT be self-approved by the generator (this exact gate wall rejected/escalated two prior consistency runs on nucleus-node/nucleus-oidc-provider). Do NOT treat `persona consistency` as your verification gate and do NOT block on it. Instead: (1) grep the crate yourself for genuine offenders — `grep -rn '\.unwrap()\|\.expect(' crates/nucleus-cli/src` for tau_panic (replace with `?` ONLY inside fns returning Result/Option; LEAVE provably-safe unwraps like on a just-inserted map key or a compile-time-constant regex), and look for unguarded blocking subprocess/socket/lock calls for gamma_bound (add a timeout/kill guard). Optionally align edition + shared dep versions to the workspace (sigma_manifest). (2) Make the SMALLEST correct change — a handful of genuine fixes beats a broad sweep. Do NOT touch any authorization/capability-guard/access-control path (δ_authz/ifc_leak are advisory, human-only). Do NOT modify .vendor-neutrality-allowlist or anything outside the crate (a prior run was rejected for touching an out-of-envelope file). VERIFY strictly with cargo, which IS runnable: `cargo build -p nucleus-cli` then `cargo test -p nucleus-cli` — both MUST stay green. Report exactly which offenders you fixed (fn + gap class) and which you deliberately LEFT as false positives with a one-line reason each. Do not claim a fix you did not make — closure is re-measured after landing.

### DIFF
--- a/crates/nucleus-cli/src/run.rs
+++ b/crates/nucleus-cli/src/run.rs
@@ -780,7 +780,16 @@ fn create_pod_via_node(
 ) -> Result<String> {
     let url = format!("{}/v1/pods", node_url.trim_end_matches('/'));
     let body = serde_yaml::to_string(spec)?;
-    let mut request = ureq::post(&url).header("content-type", "application/yaml");
+    // Bound the request: the top-level `ureq::post` helper uses a default agent
+    // with no timeout, so a hung/unresponsive node would block the CLI forever.
+    // Mirror the 30s global timeout used by node.rs::create_agent().
+    let agent: ureq::Agent = ureq::Agent::config_builder()
+        .timeout_global(Some(Duration::from_secs(30)))
+        .build()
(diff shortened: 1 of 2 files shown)


brandon-coproduct:
Superseded: this failed Rustfmt because the generator emitted un-fmt'd code. Fixed at the source — the generator now runs cargo fmt before commit. The pilot will re-attempt this crate cleanly. Closing to unjam.

PR #1975, opened 2026-07-06
De-vendor crate `nucleus-tool-proxy` ONLY — c5-vendor-neutrality

brandon-coproduct:

De-vendor crate `nucleus-tool-proxy` ONLY — c5-vendor-neutrality. Offender files: crates/nucleus-tool-proxy/Cargo.toml, crates/nucleus-tool-proxy/src/mtls.rs, crates/nucleus-tool-proxy/tests/mtls_integration.rs, crates/nucleus-tool-proxy/tests/red_team_harness.rs. Apply the same proven RENAME/EXTRACT pattern that landed for nucleus-cli and nucleus-identity: for each vendor string in product code, either (a) rename to a neutral constant / extract behind a neutral trait/adapter, or (b) if the reference is legitimately integration-specific (e.g. a red-team test fixture asserting on a real vendor tool name), add that exact path to .vendor-neutrality-allowlist with a one-line rationale. Do NOT touch authorization/capability-guard/mTLS trust logic — only rename identifiers/strings, never change auth behaviour. Touch ONLY files under crates/nucleus-tool-proxy/ (plus the repo-root .vendor-neutrality-allowlist if you allowlist a path). VERIFY (must stay green): `cargo build -p nucleus-tool-proxy`, `cargo test -p nucleus-tool-proxy`, and `bash ci/no-vendor-strings.sh` (the closure is re-measured on landing — the offender count for this crate MUST drop to 0). Report which files you neutralized vs. allowlisted, with the rationale for each allowlist entry.

### DIFF
--- a/.vendor-neutrality-allowlist
+++ b/.vendor-neutrality-allowlist
@@ -1,0 +1,13 @@
+# .vendor-neutrality-allowlist
+#
+# Paths listed here are exempt from the vendor-neutrality closure. A path
+# belongs here only when its vendor reference is legitimately integration-
+# specific — an opt-in test fixture that drives a real vendor API and cannot
+# be neutralized without destroying its function. Each entry needs a one-line
+# rationale. This is a last resort; prefer rename/extract in product code.
+
+# Opt-in (feature = "red-team"), network-gated (skipped unless ANTHROPIC_API_KEY
+# is set) integration test that drives a live vendor LLM as an adversarial
+# red-team agent against Portcullis. The vendor endpoint/headers/model id ARE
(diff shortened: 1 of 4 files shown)


brandon-coproduct:
Superseding: this de-vendor PR went DIRTY from the shared-branch collision. Fixed at the source — de-vendor now uses per-crate branches (fix/vendor-neutrality-<crate>) and the pilot's disjointness guard now reserves a crate through its open-PR window, so the crate will be re-attempted cleanly without churn. Closing to clear the scope-hygiene tangle.

PR #1979, opened 2026-07-06
De-vendor crate `nucleus-node` ONLY (c5-vendor-neutrality) — this is NOT…

brandon-coproduct:

De-vendor crate `nucleus-node` ONLY (c5-vendor-neutrality) — this is NOT c20-consistency; do NOT invoke `persona consistency`. Touch ONLY crates/nucleus-node/ (the single flagged file crates/nucleus-node/src/main.rs). For each vendor token: prefer extracting behind a neutral abstraction or renaming to a generic term that PRESERVES the original meaning (e.g. 'LLM_API_TOKEN', 'AI agent', generic policy names per CLAUDE.md). Do NOT destroy a canonical upstream URL, a concrete CI/workflow path, or genuinely-generic build/host infra to satisfy a grep — if a reference is load-bearing and generic, add its path/token to .vendor-neutrality-allowlist with a one-line written rationale instead of mutilating it. Before touching anything, confirm the actual c5 gate (ci/no-vendor-strings.sh) is what flags it, and re-run that gate plus `cargo check -p nucleus-node` after the change.

### DIFF
--- a/crates/nucleus-node/src/main.rs
+++ b/crates/nucleus-node/src/main.rs
@@ -347,7 +347,7 @@ struct FirecrackerPod {
 ///
 /// The container image can be either:
 ///   - `nucleus-tool-proxy:latest` (proxy mode: audit + policy enforcement)
-///   - A Claude Code CLI image like `gt-executor:latest` (direct mode)
+///   - An AI agent CLI image like `gt-executor:latest` (direct mode)
 ///
 /// The mode is determined by the PodSpec label `nucleus.io/proxy-mode`.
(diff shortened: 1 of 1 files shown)


brandon-coproduct:
Superseded — `crates/nucleus-node` is already vendor-neutral on main (the claude/user-path tokens this PR removed were de-vendored by a later change; a focused vendor-token scan is clean). This stale branch now only conflicts and locks the crate's blast-radius from the dispatch loop. Closing to free it; no work is lost.

PR #1984, opened 2026-07-06
De-vendor crate `nucleus-econ-kernels` ONLY (c5-vendor-neutrality)

brandon-coproduct:

De-vendor crate `nucleus-econ-kernels` ONLY (c5-vendor-neutrality). Envelope: crates/nucleus-econ-kernels/ — the single flagged file is crates/nucleus-econ-kernels/src/vcg.rs. Replace the vendor name with a neutral term (generic LLM/AI-agent/orchestrator phrasing per CLAUDE.md), OR if the reference is load-bearing add crates/nucleus-econ-kernels/src/vcg.rs to .vendor-neutrality-allowlist with a one-line rationale. Touch ONLY files under crates/nucleus-econ-kernels/ (plus the allowlist file if that path is taken). VERIFY in-worktree: `cargo build -p nucleus-econ-kernels` stays green, and `grep -rn <vendor> crates/nucleus-econ-kernels/` shows the name is gone (or only in the allowlisted path) — paste both. Do NOT run any `ci/` script and do NOT block on the no-vendor gate; that runs in nucleus CI post-landing. A run is a fix if the vendor ref is removed/allowlisted and the crate builds. No scratch tooling, .gitignore, or stray files in the committed diff.

### DIFF
--- a/crates/nucleus-econ-kernels/src/vcg.rs
+++ b/crates/nucleus-econ-kernels/src/vcg.rs
@@ -116,8 +116,7 @@ pub struct IntegerBid {
 /// a `value_micro_usd` field that the welfare computation summed; that
 /// design admitted a hidden two-utility-function bug (IR could fail
 /// whenever bid ≠ proposal value, which is precisely when separate
-/// proposal values exist). The Week 6 Option A fix from
-/// `/Users/bcrisp/.claude/plans/let-s-web-search-look-modular-rivest.md`
+/// proposal values exist). The Week 6 Option A design fix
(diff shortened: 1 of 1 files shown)


brandon-coproduct:
Still-valid de-vendor work, but this branch conflicts with main (main de-vendored `vcg.rs` divergently). The crate still has vendor tokens, so c5-vendor-neutrality will re-propose and re-dispatch it cleanly against current main. Closing this stale branch to retire the stuck accept and let the loop redo it conflict-free.

PR #1994, opened 2026-07-07
De-vendor crate `ctf-engine` ONLY (c5-vendor-neutrality)

brandon-coproduct:

De-vendor crate `ctf-engine` ONLY (c5-vendor-neutrality). Envelope: crates/ctf-engine/ — 2 flagged source files: crates/ctf-engine/src/level.rs, crates/ctf-engine/src/takeaways.rs. For each vendor name in product code, extract it behind a neutral term/const/trait (prefer renaming to a generic 'LLM'/'AI agent' equivalent — this is CTF educational content, so the vendor name is almost certainly narrative text, not a real dependency). Do NOT touch .vendor-neutrality-allowlist (adding paths there is unauthorized for this envelope and has been rejected on prior runs — a config file outside crates/ctf-engine/). Touch ONLY files under crates/ctf-engine/. VERIFY in-worktree: `cargo build -p ctf-engine` stays green AND `grep -rin <vendor> crates/ctf-engine/` returns nothing. Do NOT run any ci/ script — the authoritative no-vendor gate runs in nucleus CI post-landing; it is not your local gate.

### DIFF
--- a/crates/ctf-engine/src/level.rs
+++ b/crates/ctf-engine/src/level.rs
@@ -86,7 +86,7 @@ impl Level {
     /// Construct an agent-safe variant of a level.
     ///
     /// Identical lattice logic but with sanitized filesystem content that
-    /// won't trigger host safety layers (OpenAI content filters, etc.).
+    /// won't trigger host safety layers (LLM provider content filters, etc.).
     pub fn new_agent_safe(number: u8) -> Self {
         let mut level = Self::new(number);
         level.filesystem = Self::agent_safe_filesystem();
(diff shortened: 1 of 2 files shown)


brandon-coproduct:
Superseded — `crates/ctf-engine` is already vendor-neutral on main (the claude/user-path tokens this PR removed were de-vendored by a later change; a focused vendor-token scan is clean). This stale branch now only conflicts and locks the crate's blast-radius from the dispatch loop. Closing to free it; no work is lost.

PR #2005, opened 2026-07-09
fix(nucleus-externality): clippy for_kv_map — unblock the merge queue

brandon-coproduct:
## The problem
CI runs `cargo clippy --all-targets --all-features -- -D warnings`. **Rust 1.97.0 newly fires `clippy::for_kv_map`** on `nucleus-externality/src/profile.rs:117` (a toolchain-drift CI break — the lint was tightened, not the code changed). Under `-D warnings`, that one lint fails Clippy on **every open PR**, jamming the entire merge queue — which is why 0 PRs have landed recently.

## The fix
`for (_dim, claim) in p.dimensions.iter()` uses only the value → `for claim in p.dimensions.values()`. Same deterministic `Ord`-order iteration over the `BTreeMap`, **byte-identical canonical digest**. One line.

## Verified
`RUSTFLAGS="-D warnings" cargo clippy -p nucleus-externality --all-targets --all-features` is clean (the CI log flagged only this one site). Unblocks #2003, #2002, and the rest of the queue.

### DIFF
--- a/crates/nucleus-externality/src/profile.rs
+++ b/crates/nucleus-externality/src/profile.rs
@@ -113,8 +113,8 @@ pub fn canonical_externality_bytes(p: &ExternalityProfile) -> Vec<u8> {
     out.extend_from_slice(PROFILE_DOMAIN);
     let n = p.dimensions.len() as u32;
     out.extend_from_slice(&n.to_be_bytes());
-    // BTreeMap's iter is in `Ord` order, which is deterministic.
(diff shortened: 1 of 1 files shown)


brandon-coproduct:
Superseded by #2010 (rust-toolchain.toml pin to 1.96.1), which fixes the clippy::for_kv_map drift at the root — the toolchain no longer floats to 1.97, so the per-site fixes here are unnecessary. Closing in favor of the declared-input fix.

PR #2087, opened 2026-07-28
fix(effects): bind discharge to its action — a read bundle could authorize a write

brandon-coproduct:
## The bug

Every mediated method on `NucleusRuntime` took the discharge bundle **by value and never inspected it**. `require_scope` was not called anywhere in `runtime.rs`. The bundle proved only that *a* preflight ran — never that a preflight ran for **this** action.

Confirmed before fixing, not inferred:

```rust
let bundle = rt.preflight_read()?;        // ReadFiles / AuditLogAppend
rt.write_file(&path, b"...", bundle)?;    // returned Ok(()) — and wrote the file
```

This is exactly the confused deputy that the `operation` / `sink_class` fields on `DischargedBundle` were added to prevent. The fields existed; nothing read them.

## Severity — precisely

**Not a total bypass.** `PolicyEnforced` still applied the coarse lattice gate on every call, so the capability had to be non-`Never`. What a substituted bundle defeated is the **eight obligations**: integrity gate, path allowed, derivation clear, no adversarial ancestry, budget, delegation ceiling, task scope, inputs authorized.

The exploitable shape: *an action whose own discharge would fail is performed under a discharge that succeeded for something cheaper.* A session too tainted to push — `NoAdversarialAncestry` would reject it — could push on a read bundle.

## A second, independent gap

`check_path_allowed` ran only inside `preflight_write`. So `write_file` trusted that the preflight had checked — but not necessarily **this** path. The path is not part of the scope, so a bundle honestly earned for an allowed path redirects anywhere and stays in scope:

```rust
let bundle = rt.preflight_write("/allowed/dir/x")?;   // legitimately earned
rt.write_file("/somewhere/else", b"...", bundle)?;    // same scope, different target
```

Established by **perturbation**: removing the re-check makes `a_write_bundle_for_one_path_does_not_authorize_a_write_to_another` fail, so the test depends on that line and nothing else.

## The fix

- New `require_scope_for` helper; all seven mediated methods check the bundle against the same `(Operation, SinkClass)` pair their `preflight_*` builds.
- `write_file` re-runs `check_path_allowed`.
- New `RuntimeError::ScopeMismatch`, naming both the authority held and the action attempted.

No new machinery — `require_scope` already existed as `pub(crate)` in the same crate.

## Docs corrected

`verified-claims.md` claim 6 said there is no way to execute a side effect without passing discharge. True as written, and misleading: you had to pass *a* bundle, not the *right* one. It now separates what the type system establishes from what the runtime check carries, and records the two complete-mediation properties still open.

`production-delta.md`'s enforcement-completeness row cited 146 call sites in `nucleus-claude-hook` — a crate that no longer exists. Restated against the effect layer, which is where the real gap is:

> 10 of 13 effect-trait methods — including `FileEffect::write` — take no `DischargedBundle` at all and are gated only by the coarse lattice. The 3 that do take one receive it by reference, so a correctly-scoped bundle can be replayed.

That is the next piece of work, not this PR.

## Verification

- 88 lib tests + 16 doctests pass.
- `cargo clippy -p portcullis-effects --all-targets -- -D warnings` clean.
- `cargo check --workspace` clean.
- Vendor gate clean.
- Red→green recorded: both new tests fail on the parent commit for the right reason (`expect_err` receiving `Ok`), pass after.

**Pre-existing unrelated breakage:** `cargo check --workspace --all-targets` fails in `portcullis-core/tests/provenance_e2e.rs` (`no method named content_hash`). Confirmed present on a clean `origin/main` worktree — not from this change, and not fixed here.


brandon-coproduct:
Superseded by #2090, which contains this work plus everything stacked on it — verified by tree diff: nothing in this branch is absent from #2090.

Closing rather than fixing this branch's failing Rustfmt check, since two paths to `main` carrying overlapping content is the thing that already caused a problem here: auto-merge collapsed the original stack in an order that stranded #2089's refactor off the merge path. #2090 is the single branch that carries the whole arc:

- `fix` — bind discharge to its action (the live confused-deputy bug)
- `proof` — scope binding, machine-checked over the Aeneas-extracted Rust
- `refactor` — all 13 effect methods take an `Authority` by value
- `feat` — every mediation decision witnessed in a hash chain
- `style` — rustfmt

#2090 currently has no failing checks.

PR #2089, opened 2026-07-28
refactor(effects): FileEffect takes an Authority by value — scoped and one-shot

brandon-coproduct:
**Stacked on #2088** (which is stacked on #2087). Review bottom-up.

First trait of the B1/B2 cutover, staged per the plan.

## What was wrong

`FileEffect::write` is the most security-relevant method in the system, and it sat in the **lattice-only tier**: gated by `write_files != Never` and nothing else. None of the eight obligations — integrity gate, path allowed, derivation clear, no adversarial ancestry, budget, delegation ceiling, task scope, inputs authorized — reached a file write.

## What changed

All four methods (`read`/`write`/`append`/`glob`, plus `read_str` and the async mirror) take `Authority` **by value**. Two properties follow:

**Scoped.** The authority names the `(Operation, SinkClass)` it was earned for; `PolicyEnforced` spends it at the gate. A read authority cannot pay for a write.

**One-shot.** By value means moved — replay is a move-after-move *compile* error, not a policy someone has to remember. This closes the seam #2084 named explicitly: `spend()` handed the bundle back and ambient `&DischargedBundle` resumed. `PolicyEnforced` now re-wraps into a fresh `Authority` before delegating, so the property survives the last hop instead of decaying there.

## Two testing notes worth reading

**The scope test is checked against the right cause.** `the_scope_refusal_is_not_the_lattice_gate` sets `write_files: Always`, so if it passed for the wrong reason the coarse capability would be doing the work while the obligations still weren't. Without that, "write refused" proves nothing.

**The `compile_fail` test was not running.** I first wrote it inside `#[cfg(test)] mod tests`, where rustdoc does not collect it — caught because the doctest count stayed at 16. Moved onto the public `FileEffect` trait (now 17), then perturbed: deleting the replay line makes the snippet compile and the test fail. So the failure depends on the replay and nothing else — checked, not assumed, per the discipline `Authority`'s own doctest documents.

## Blast radius: contained

No crate outside `portcullis-effects` calls `FileEffect`. The Python binding only references `read_file` in docstrings. `NucleusRuntime`'s public signature is **unchanged** — it still takes a `DischargedBundle` and wraps it internally, so the ergonomic path is untouched.

Its `require_scope_for` check is kept alongside the new effect-level one: the runtime yields the precise `RuntimeError::ScopeMismatch`, and the effect layer covers callers that bypass the runtime.

## What remains

Docs now carry the real count. **6 of 13 methods are still unmediated** — `WebEffect::{fetch,search}`, `ShellEffect::run`, `GitEffect::{commit,push}`, `AgentSpawnEffect::spawn` — and 3 take the bundle by reference, so replay is still possible there. `production-delta.md` moves from Open to **Partial**, not Done.

## Verification

- 91 lib tests + 17 doctests pass (`--all-features`).
- `cargo clippy -p portcullis-effects --all-targets --all-features -- -D warnings` — clean.
- `cargo check --workspace` — clean.
- `--features async` builds.


brandon-coproduct:
Pushed a second commit (`74a2f85b`) completing the B1 surface — **all 13 effect-trait methods now require an obligation token, up from 3.**

This converts the remaining six that took none at all: `WebEffect::{fetch,search}`, `ShellEffect::run`, `GitEffect::{commit,push}`, `AgentSpawnEffect::spawn`, plus async mirrors.

## `GitEffect::push` is the one that mattered

Highest-consequence sink in the system, previously gated by `git_push != Never` and nothing else — so a session too tainted to discharge for `GitPush` could still push. That is exactly the escalation `scope_admits_no_escalation` names on the Lean side in #2088; it is now unreachable in Rust.

All six sink pairings were checked against `operation_allowed_for_sink` before being written rather than guessed — each is one of the 27 structurally earnable pairs, so a preflight can actually produce them.

## Three tests, and what makes them mean something

- `a_commit_authority_will_not_pay_for_a_push` — the headline. Uses `RecordingEffects` as the inner impl, because a correctly-scoped push through `RealEffects` would run a real `git push`.
- `every_mediated_method_refuses_an_out_of_scope_authority` — same property across all nine, asserting the refusal precedes the inner effect (recorder stays empty), not merely that *an* error came back.
- `every_mediated_method_accepts_its_own_authority` — the counterpart. A scope check that refused everything would pass the test above and break the runtime; this pins all ten correctly-scoped calls reaching the effect.

Every capability is `Always` in these tests, so a pass cannot come from the coarse lattice doing the work while the obligations still don't.

**Perturbation:** removing the `spend_for` call in `push` fails both scope tests, and the cross-surface one names `push` specifically.

## Updated scorecard

| | count | replayable? |
|---|---|---|
| `Authority` by value | **10** | no — compile error |
| `&DischargedBundle` | 3 | **yes** |
| no token at all | **0** (was 10) | — |

The three by-reference holdouts are `ShellEffect::run_argv`, `AsyncShellSpawnEffect::run_argv_async`, `NetEffect::fetch`. Unlike everything in this stack so far, they have callers *outside* `portcullis-effects` (`nucleus/src/command.rs`, `nucleus-tool-proxy`), so they're their own unit of work rather than a tail on this one. Both docs say so explicitly.

Verification: 94 lib tests + 17 doctests, clippy clean under `-D warnings` with `--all-targets --all-features`, workspace check clean with and without `--all-features`.

PR #2095, opened 2026-07-28
feat(lint): `mediated` — a call-graph-closed check that discharges the Lean premise

brandon-coproduct:
Task #5. Stacked on #2094. This is part **B** of the global-mediation factoring.

## Why a lint at all

Global mediation is a whole-program property, so **Lean cannot prove it directly**: the effect layer does real I/O (outside the Charon/Aeneas subset), and "every effect anywhere" quantifies over code not yet written.

**seL4 has the same shape and resolves it the same way** — it assumes compiler/assembly/hardware correctness and imposes *syntactic* restrictions (notably: no calls through function pointers) checked outside Isabelle, so the proof has a statically known call graph.

| Part | Mechanism | Establishes |
|---|---|---|
| A | Lean theorem over the extracted model | *if* every effect entry point consumes a scoped `Authority`, *then* every trace is mediated |
| B | **this lint** | the premise of A, over the real Rust |
| C | trusted, and stated | rustc's affine moves; deny-set completeness; extraction fidelity |

## Deliberately the opposite posture from its sibling

`nucleus-guarantee-lint` is careful that `aeneas_eligible` is *"a SCREEN, not a proof — a clean pass asserts nothing"*. Correct there, **wrong here**: a clean pass must assert something real or it cannot discharge a premise. Separate crate, and both READMEs now say which is which.

## Verified by running it, not by reading it

`check_fn` sees one function at a time, so the closure is a least fixpoint at `check_crate_post`. Eight fixtures, all confirmed via `cargo dylint`:

- transitive reach **two hops deep is caught** — a per-function lint misses this entirely;
- a boundary on the path **absorbs** propagation, so the public caller is clean;
- **`&Authority` is not a boundary** — it's the replayable ambient shape the affine cutover removed; without this case a regression to it reports clean;
- calls through a function pointer are **reported**, not skipped.

**Perturbation:** deleting the fixpoint loop drops `public_two_hops` and nothing else, so the transitive case depends on the closure rather than on incidental direct-I/O detection.

## I got a design decision wrong and corrected it mid-flight

The first version reported every cross-crate call as an unfollowable edge. Running it produced **299 such reports against 15 real findings** on `portcullis-effects` alone — and the reasoning was wrong, not just noisy. Dylint runs per crate, so a cross-crate callee is analysed by *its own* pass, and it must be exported to be callable, so `is_exported` can't hide it. The finding surfaces where the defect is. Calls *out* of the mediated set rest on deny-set completeness, which is already stated as load-bearing — reporting them double-counted an assumption. Removed, with the argument recorded at the call site.

## First findings (not acted on here)

15 functions reach raw I/O with no `Authority`, splitting exactly along the "mediated set" boundary this forces us to decide:

- **Agent-facing, worth closing:** `capability_traits::{read_file, write_file}` — unmediated filesystem helpers sitting next to the mediated effect traits.
- **Infrastructure, probably out of set:** `*::load_from_dir` config loaders, `sink::JsonlSink`, `merkle::MerkleSink`. Runtime's own I/O, same category as jailer spawn.

## Honest gaps

**UI tests are `#[ignore]`d.** On macOS the dylib resolves `rustc_driver` via `@rpath`, which `DYLD_FALLBACK_LIBRARY_PATH` doesn't satisfy; `DYLD_LIBRARY_PATH` does (direct `dlopen` succeeds) but SIP strips it from spawned children. **Pre-existing** — the sibling lint fails identically on this host, which I verified before blaming my own crate. **No `ui/main.stderr` is committed**, because generating one requires running the harness; a hand-written baseline would be fabricated. The reason lives in the test file, not just the README.

**A clean pass proves the authority is *demanded*, not *spent*** — `drop(authority)` passes. What closes that is elsewhere (`!Clone`, `#[must_use]`, `spend` as the only consuming method, and the Lean theorem over the scope check).

Also: the first push swept in `target/` (2,600 files, two 95 MB rlibs). Second commit untracks it and adds the `/target` .gitignore the sibling crate always had.


brandon-coproduct:
Pushed the **mediated-set decision** (task #6) on top of the lint — `docs/architecture/mediated-set.md`.

The guarantee is relative to a boundary, so the boundary had to be written down before the Lean theorem's premise means anything. The set: `portcullis-effects`, `nucleus` (`Sandbox` + `Executor`), `portcullis-core::capability_traits`.

The distinguishing question is **not** "does it touch the disk" but **on whose authority** — an effect performed because an agent asked is in; the runtime doing its own job is out, because there is no agent authority to discharge against.

## Three holes the lint found, now closed

**`capability_traits::{read_file, write_file}`** performed real `std::fs` calls behind only a phantom-type check — no discharge, no `FlowTracker`, no path allowlist — and were reachable from `NucleusRuntime::with_typed_context`, **whose own doctest demonstrated reading a file that way**. Six of the eight operations in that module were already stubs saying "use the mediated path"; these two silently weren't. Now stubs matching their siblings. No production callers.

**`Sandbox::read_to_string_for_search`** took no `DecisionToken` and carried a comment saying the caller *"must have already obtained a GrepSearch decision"* — a convention where an enforcement belongs, on an agent-reachable read. Now requires an `Authority` scoped to `(GrepSearch, AuditLogAppend)`.

## A cost I should name rather than let you discover

The search loop now mints **one discharge per file** — a grep over N files runs N preflights. That is the affine model working as designed (an `Authority` buys one read), and the alternative — one token covering a whole directory walk — is exactly the replay the by-value cutover removed. If it ever matters, the answer is a coarser sink class, not a reusable token.

## Exclusions argued, not asserted

Eighteen remain excluded. Two entries are deliberately uncomfortable:

- **The audit sinks** are called out as *the weakest argument on the page*: an unmediated write path carrying attacker-influenced content. What bounds it is that the destination is fixed at construction, and that requiring an agent authority to write the audit log would let a *denied* agent suppress its own denial record — the cure would be worse.
- **`Sandbox::{exists, exists_approved}`** are listed as **signature-blocked, not judged safe**. They return `bool` with no error channel, so an authority is a signature change rather than a check. `exists` is an information-disclosure probe and belongs in the set by rights.

Lint re-run confirms `capability_traits` no longer appears. Tests green across all four affected crates, clippy clean under `-D warnings`, `cargo fmt --all -- --check` clean.

brandon-coproduct:
Pushed **task #3 — the trace half of the proof.** Four new theorems over the Aeneas-extracted state machine, joining the four scope-predicate ones:

| Theorem | Says |
|---|---|
| `effect_requires_held` | no authority, no effect |
| `effect_consumes_the_authority` | success returns the machine to idle |
| `refusal_preserves_the_authority` | a denial does **not** burn the token |
| `no_replay_without_a_fresh_discharge` | the two-step corollary |

**All eight carry `[propext, Classical.choice, Quot.sound]`** — no `sorryAx`, no Aeneas `*External`.

## Proved over generated code, not a hand model

`MedState` / `MedAction` / `med_step` live in `extracted/mediation.rs` and go through `charon --start-from` → `aeneas`, exactly as `scope_admits` did. **The kill criterion did not fire a second time.** The machine uses a flat `held: bool` rather than a payload-carrying enum, so no discriminated union lands in the proof's critical path.

`refusal_preserves_the_authority` is the one I'd flag: a denial that *consumed* the token would be a denial-of-service on the honest path — a wrong-scope attempt could burn a legitimate authority. Worth proving rather than assuming.

## The whole factoring now exists

| Part | Mechanism | Status |
|---|---|---|
| A | Lean, over the extracted model | ✅ 8 theorems, clean axioms |
| B | `mediated` Dylint pass, call-graph-closed | ✅ this PR |
| C | trusted, and named | rustc's affine moves; deny-set completeness; extraction fidelity |

The file header states plainly what Lean **cannot** do here: "every effect entry point consumes an Authority" is a whole-program property over an open program, so there is nothing to quantify over. These theorems are the *conditional*; the lint discharges the premise. seL4 has the same shape.

It also corrects a stale caveat — the header still claimed 10 of 13 effect methods took no bundle, which has been false since #2090.

## Rust parity

Three exhaustive machine tests: no effect succeeds from idle (247 pairs); an authority pays for exactly one effect with the replay refused (247); and a refusal never consumes, swept over every earned × attempted pair — **61,009 cases**.

`med_step`/`med_idle` added to `EXTRACT_ROOTS`; twin workflows verified to still mirror.

194 `nucleus-ifc-kernel` tests pass, `lake build` completes (1686 jobs), `cargo fmt --all -- --check` clean.

brandon-coproduct:
Pushed **task #2 — the Merkle transparency log**, replacing the linear hash chain. Merkle now rather than signed-linear-first, per your call.

## What the chain could not do

A linear chain detects reordering and edits. It cannot express either proof a transparency log exists to provide:

- **Inclusion** — "this effect is in the log", in `O(log n)`. Without it, an auditor checking one action must receive every action.
- **Consistency** — "the log I see now only grew from the log I saw before." This is the one that matters: **an operator who discards a suffix and recomputes produces a chain that verifies perfectly.** There was nothing for the old structure to catch.

`a_rewritten_log_cannot_prove_consistency_with_an_earlier_checkpoint` is that attack end to end — keep 5 entries, append 3 different ones, land on the **same size** so the sizes can't betray it — and asserts that no proof at any size bridges the auditor's checkpoint to the rewritten log.

## Faithful to the spec, including the parts that are easy to skip

The `0x00`/`0x01` leaf/node prefixes are load-bearing — without them a leaf and an internal node can hash identically (the classic Merkle second-preimage). Pinned by its own test.

Inclusion is checked at **every size 1–33 for every leaf**, so the unbalanced non-power-of-two trees, where proof lengths differ per leaf, are covered rather than assumed. Consistency likewise for every prefix up to 17. And `a_forged_leaf_has_no_inclusion_proof` covers the direction that matters: a verifier that accepts anything proves nothing.

## A bug worth recording

My first `verify_consistency` hashed only the *second* root on a right-border node. Every prefix then failed to reproduce `root_m`, and the exhaustive test caught it immediately. The corrected code hashes **both**, with a comment saying why — the asymmetry looks plausible, which is exactly why it needed the test rather than a reading.

## Signed checkpoints

Ed25519 (already a dependency) over fixed-width `size || root`. Tampered root, tampered size, and foreign key all covered.

## Still not established — stated in the module, not implied away

- **Nothing is published.** A signed checkpoint proves the operator *asserted* a root, not that anyone else saw it. A **forked** log — two validly signed checkpoints of the same size with different roots — needs a witness to detect. Rekor anchoring is the escalation path; not done.
- The signing key is caller-supplied and **not bound to a measured binary**. That is the attestation work.
- Coverage remains the ten methods `PolicyEnforced` gates.

The completeness test in `lib.rs` now verifies each of its ten receipts by inclusion proof against the root, rather than walking a chain — the check an auditor actually runs.

106 lib tests + 17 doctests, clippy clean under `-D warnings`, workspace check clean, fmt clean.

---

**All six queued tasks are now done.** The remaining honest gap in the whole arc is attestation: binding a checkpoint to a *measured* policy-kernel binary, so the receipt says which verified monitor produced it.

PR #2098, opened 2026-07-29
harden(egress): extract the CIDR/port matcher for the confinement proof

brandon-coproduct:
Groundwork for the egress-confinement theorem, plus a finding.

## Why egress is where the proof has to land

Complete mediation on the effect layer bounds what an agent may **ask** for. It does not bound what a shell does once started: one `(RunBash, BashExec)` authority buys arbitrary syscalls inside the guest, and `bash -c 'curl … | sh'` never passes through `NetEffect::fetch`. The receipt records `RunBash / Allowed` — one line, unbounded consequence.

So for shell effects the real enforcement boundary is the network policy. Given autonomous jailbreak agents now reported around **97%** success, the defence cannot be preventing the jailbreak; it has to be that a jailbreak is non-decisive. That claim currently rests on iptables rules nobody has proven anything about.

## What this lands

`crates/nucleus-ifc-kernel/src/extracted/egress.rs` — `in_cidr`, `rule_matches`, `rule_admits` — added to the scoped Charon/Aeneas extraction roots so CI generates and drift-checks the Lean definitions.

**Scalar-only, matching every other module in `extracted/`.** Aeneas emits derived comparisons and most collection operations as opaque axioms, which would put unspecified assumptions on the proof’s critical path. So the fold over a rule list is deliberately *not* extracted: the list semantics (first match wins, default `DROP`) belongs in Lean over `List Rule` with this matcher as its primitive. That is the same split the [verified iptables work](https://link.springer.com/article/10.1007/s10817-017-9445-1) uses — formal semantics over rule lists in the prover, matchers underneath.

The cost is an explicit correspondence obligation, stated in the module docs: the Lean fold must be the order `net.rs` actually appends rules in. That will be discharged by a test, not a proof — the same posture as the `mediated` lint.

The matcher is where an off-by-one silently widens an allowlist, so it is tested against an **independent mask formulation across all 33 prefix lengths**, plus block boundaries, out-of-range prefixes (which narrow rather than widen), port-specific vs port-agnostic matching, and that a matching `DENY` rule matches but never admits.

## No Lean theorems in this PR, deliberately

Aeneas **cannot run on macOS** — this workflow is documented Linux-only (`aeneas-ifc-scoped.yml:18`), because the local charon-ml build is broken. The generated defs therefore do not exist on my machine, and theorems stated over guessed generated names could not be built or checked here. Writing unbuildable proofs would be worse than writing none. Once CI generates the defs from these roots, the theorems have real names to be stated over.

## A finding, recorded but not fixed here

`parse_rules` accepts whatever CIDR a PodSpec supplies. **`allow: ["0.0.0.0/0"]` is a nominal network policy that permits unrestricted egress**, and nothing rejects or warns about it.

This is the same class as the `pci=off` gap in #2097: spec input dissolving a hardening floor. It is not fixed here because whether open egress is ever legitimate is a policy call rather than a bug fix — but `the_zero_prefix_matches_everything` pins the current behaviour so any fix has a test to change.

Also observed while reading: the ruleset is built by sequential `iptables` subprocess invocations rather than existing as a value. Nothing can be proven over it in that form. Making it a pure value with a thin imperative applier — the shape `NetnsPlan::decide` already has in the same file — is the refactor the confinement theorem needs next.

## Verification

- `cargo test -p nucleus-ifc-kernel` — 198 + 9 + 1 + 1 green; 7 new tests.
- clippy + fmt clean.
- Extraction-root workflow and its noop twin still mirror each other exactly (checked programmatically).


brandon-coproduct:
## Correction: Aeneas runs fine on macOS

The claim in this PR’s description — *"Aeneas cannot run on macOS"* — **is wrong.** I took it from this workflow’s own comment instead of testing it. Pushed a commit that proves the theorems and fixes the comment.

On macOS 15 / aarch64, `charon 0.1.223` + `aeneas d71d2e3` from `~/.cargo/bin`: scoped extraction of `extracted::egress` runs clean, **4/4 functions translated, 0 opaque**.

The real gotcha, which is what a stale build looks like from the outside:
- `charon cargo` **exits 0 and emits nothing** when cargo has nothing to recompile → `cargo clean -p nucleus-ifc-kernel` first.
- The `.llbc` lands at the **repo root**, not the crate directory. My first `find` used the wrong cwd and I concluded "no output" when the file was there.

The stale comment is replaced with both facts, so the next person does not lose the same hour.

## So the theorems landed

Eight, over the generated defs, all `[propext, Classical.choice, Quot.sound]`, zero `sorry`:

- **`unmatched_is_dropped`** — the confinement theorem: if no rule admits a destination, the verdict is Drop.
- **`accept_implies_a_granting_rule`** — contrapositive; passage is always attributable to a named rule.
- **`deny_before_allow_wins`** — an earlier Deny beats a later Allow, which is what `net.rs`’s append order buys under first-match-wins.
- **`rule_admits_implies_allow`** — a DENY rule never admits, however well it matches.
- **`zero_prefix_matches_everything`** — `0.0.0.0/0` admits everything, stated as a theorem rather than a footnote.
- Plus four totality lemmas — the generated defs live in Aeneas’s `Result` monad, so without them every other theorem would be vacuous on a `fail` branch.

**`in_cidr` was rewritten** from `net >> (32 - prefix)` to an explicit mask table. In the `Result` monad a `u32` subtraction is *fallible*, so the shift version put an underflow obligation on every downstream theorem. `&&&` and `==` cannot fail, so the matcher is now **total by construction**. The table is checked against the independent shift formulation across all 33 prefix lengths.

**`native_decide` was tried and rejected** for `zero_prefix_matches_everything` — it would have added `Lean.ofReduceBool`, and a kernel-external evaluation on the critical path is exactly the assumption this file exists to avoid. Replaced with a real proof.

## Perturbation

Dropping `&& rule.allow` from `rule_admits` **in the Rust**, re-extracting, and rebuilding fails `rule_admits_implies_allow` and `unmatched_is_dropped`, with `sorryAx` appearing in the axiom audit. The theorems constrain the code, not a model of it.

## Still not discharged

That the Lean fold’s order is the order `net.rs` actually appends in. That needs the ruleset to be a **value** rather than a sequence of `iptables` subprocess calls — the refactor I flagged, now the blocking next step rather than a nice-to-have. Nor that iptables implements these semantics. Nor anything about DNS-tunnelled exfiltration to an *allowed* destination: confinement is about where packets may go, not what they carry.

brandon-coproduct:
## The correspondence gap is closed

The theorem folds a list of rules. **The code had no list** — the ruleset existed only as a sequence of `iptables` subprocess calls spread across two filtered loops, so there was nothing for the fold to correspond *to*. The theorem was about an intention.

`net::egress_chain` now returns the chain as an **ordered value**, and `apply_host_policy` applies exactly that list in exactly that order, in **one pass**. Two filtered passes would re-encode the deny-before-allow ordering in the applier as well as the builder, and the two copies could drift — which is the kind of gap the proof is meant to close, not depend on.

`model_rule` / `model_chain` convert into the extracted `egress::Rule` the theorem is stated over, so the correspondence targets the proof’s own types rather than a restatement.

## Two deliberate choices

**`model_chain` returns `None` for IPv6** rather than converting lossily. "Outside the model" must never be reported as "covered and fine". It is all-or-nothing: a partially-modelled chain would licence a confinement claim over rules the model never saw.

**The Rust mirror of `verdict` uses the same extracted matcher** the theorem uses. Only the fold is restated — never the matching — so the test cannot agree with the proof by reimplementing the same bug twice.

## Tests

- `deny_precedes_allow_in_the_chain`
- `a_specific_deny_beats_a_broader_allow` — `deny 10.0.0.7/32` inside `allow 10.0.0.0/8` stays denied; its neighbour stays reachable
- `a_destination_no_rule_admits_is_dropped` — `unmatched_is_dropped` on a real policy, **with a positive case** so it cannot pass vacuously
- `dns_resolved_entries_do_not_outrank_a_deny` — a resolver handing back a denied address must not re-open it
- `an_ipv6_rule_falls_outside_the_model_rather_than_being_assumed_safe`

**Perturbation:** reversing the two `extend`s in `egress_chain` fails exactly the two ordering tests and nothing else.

## What remains

The only gap left in this leg is the one no amount of Rust or Lean can close from here: **that iptables implements these semantics, and that the rules are applied at all.** Diekmann et al. needed a paper for the first half and still approximate unknown match expressions.

Also unchanged: nothing here addresses **DNS-tunnelled exfiltration to an allowed destination**. Confinement is about where packets may go, not what they carry. And the baseline rules (loopback, conntrack ESTABLISHED/RELATED) sit outside `Rule` — it has no interface or ctstate field — so the fold models the policy segment for a *new* connection to an external destination, which is the case exfiltration needs.

brandon-coproduct:
## `0.0.0.0/0` now costs something

Rather than ban it, it forfeits the workload identity.

**The argument.** A SPIFFE SVID bounds how *long* a credential is useful — short-lived, rotated, 60-minute blast window. Nothing bounded *where* it could be presented. A pod with unrestricted egress held a credential presentable to any endpoint on the internet, an attacker’s included: the temporal bound survives and the spatial one is simply absent. Egress confinement is what supplies the second bound, so issuing an identity without it is issuing something whose reach nobody can state.

This is [SPIFFE’s own model](https://arxiv.org/html/2504.14760v1) rather than a bolt-on — SVID issuance is already policy-aware, matching attested properties against registration entries. Egress confinement becomes one of those properties.

## The rule

Every `allow` entry must **either stay inside non-routable space (any breadth) or name a single host (`/32`)**. "Some internal network" is fine; "some slice of the internet" is not, unless you can say which host.

- `10.0.0.0/8`, `192.168.0.0/16` → identity kept. Breadth inside private space cannot present a credential to the internet.
- `93.184.216.34/32` → identity kept. The destination set is enumerable, which is the whole property.
- `0.0.0.0/0`, `93.184.216.0/24`, **and `93.184.216.34/31`** → forfeited.

That last one is deliberate: a /31 is two hosts and still forfeits. **Named, not thresholded** — a size cutoff would be an arbitrary line to argue about; "name the host" is a property. `dns_allow` is unaffected by construction since it resolves to `/32`s, so the ordinary "let me reach this API" case keeps its identity.

## Two design calls

**Absence of a policy GRANTS.** No policy means `NetnsPlan` still creates the netns and applies default-deny with *no allow rules* — the most confined state there is, not the least. Denying there would push operators to write a policy in order to keep an identity, inverting the incentive the gate exists to create.

**A trade, not a prohibition.** The pod still boots; it just boots without a workload API. Failing the launch would make it a ban and remove the choice you asked for.

## Verification

`workload_api_port_for` is extracted so the composition that actually enforces this is testable without booting a VM, rather than three lines reviewed by eye inside a Linux-gated function. 8 new tests, 125 green.

**Perturbation:** neutering the public-reach check fails exactly the four denial tests and leaves the three grant tests passing — so the gate is not merely refusing everything.

## Residual

This **withholds the endpoint, it does not refuse to serve.** A guest that guessed the vsock port could still reach the listener. The defence-in-depth version refuses at the serving side too, keyed on the pod’s grant; that is not implemented, and is stated in the code rather than left implied.

PR #2101, opened 2026-07-29
feat(fm-2,fm-3): CB4A separation as call-graph reachability, a CI job that runs it, and the witness guard that was never written

brandon-coproduct:
Stacked on #2100. FM-2 of the secretless-broker plan.

## The MUSTs, mechanised

CB4A states its separation of duties as prose about components:

> The component that decides "yes" (**PDP**) MUST never touch credential
> material. The component that dispenses credentials (**CDP**) MUST never make
> policy decisions.

`tools/nucleus-cb4a-lint` states them as **reachability over the call graph**.

### Why a lint, when the signatures already say this

`pdp_decide` takes no store and `cdp_fetch` takes no lattice, which closes the
*direct* case at the type level. But a signature says nothing about two hops
down — and that is the shape separation-of-duty failures actually take. Nobody
adds `CredentialStore` to the decider's arguments; they add it to something the
decider calls.

Not hypothetical. Perturbing the real `nucleus-node` so `pdp_decide` reaches a
store through two intermediate helpers — **every signature in the chain clean** —
the pass reports it at the helper. Only the closure sees it.

### Forward, not backward

The sibling `mediated` pass closes **backward**, from I/O primitives up to public
entry points. This closes **forward** from named roots. Same fixpoint machinery,
opposite direction — and the difference matters: a backward closure that stops at
a crate boundary still catches the defect, because the callee's crate gets its
own pass. A forward closure that stops at a boundary genuinely stops.

So this is **half** of FM-2, and says so. The other half is `deny.toml`'s
`wrappers` ban, which permits only the composition root to link credential
material at all. Inside that crate the lint applies; outside it the ban does.
Neither is sufficient alone.

### A clean pass means nothing until it fires

Here a silent pass has a second failure mode beyond the usual one: **if no root
matched, the closure never runs**, and the result is indistinguishable from "no
violations". So the clean result rests on three perturbations, each of which
produced a report — and the first two also establish that `pdp_decide` and
`cdp_fetch` are matched as roots at all:

| Perturbation | Crate | Result |
|---|---|---|
| `pdp_decide` → hop → helper building a `CredentialStore` | `nucleus-node` | flagged, at the two-hop helper |
| `cdp_fetch` → helper reading a `PermissionLattice` | `nucleus-node` | flagged |
| `for_request` → helper naming a policy type | `nucleus-cred-broker` | flagged |

`Operation` is deliberately **absent** from the policy set: the CDP is *told*
which operation was approved — that is the decision being communicated to it, not
the CDP making one. Including it would flag the correct design, and a lint that
fires on the intended shape gets suppressed rather than obeyed.

## And a CI job that actually runs them

Both passes were run **by hand**. A lint nobody runs is documentation with a
build step — it cannot catch the regression it was written for, because the
regression arrives in a PR and the lint runs on somebody's laptop afterwards.

**Exit status is not the signal.** `cargo dylint` exits 0 even when a pass
reports, because the lints are `Warn` — verified directly: a perturbed
`pdp_decide` produced a finding *and* exit code 0. A job trusting the exit status
would have shown green while passing any number of violations through. The check
counts reports.

`cb4a_separation` is **enforced at zero**. `mediated` is reporting-only, because
it has 15 known open findings on `portcullis-effects` and making it required
would either block unrelated PRs or force those findings to be suppressed — and a
suppressed lint is worse than an unrun one, because it looks clean.

Two failure modes that would have disabled the job *silently* are handled and
named in the workflow: dylint's `@<toolchain>-<target>` filename requirement
(absent it, "Could not find --lib" reads like a config error and invites deleting
the step that makes the check work), and rustup selecting a toolchain from the
CWD rather than `--manifest-path`.


brandon-coproduct:
### Added: FM-3 — the drift guard the docs named had never been written

`authority.rs` said `every_effect_is_witnessed` "is what stops it drifting".
**No such test existed anywhere in the repo.** The claim sat in the docs while
the completeness it asserted rested on nobody forgetting — and the module's own
history records what that costs: `run_argv`, `run_argv_async` and
`NetEffect::fetch` had ended up outside the log while every safer effect was
inside it.

The named test now exists, at the layer where the invariant lives. `witnessed`'s
docs claim *every* `PolicyEnforced` method routes its authority through the log;
that is now checked by scanning the source rather than by exercising paths,
because the property covers methods not yet written.

**Two things I got wrong, both caught rather than reasoned out:**

1. **The non-vacuity check fired first.** My scan cut the file at a banner
   comment — of which there are twenty — so it examined a prefix containing no
   authority sites and passed while checking *nothing*. Indistinguishable from a
   clean result, which is exactly how the original gap went unnoticed. The
   non-vacuity assertion now runs *before* the real one.

2. **My premise was wrong.** I believed the live path recorded nothing and added
   a `ReceiptLog` to `NucleusRuntime`. Effects *are* recorded: `PolicyEnforced`
   owns a log and attaches it at every gate — and `witnessed` **replaces**
   whatever log an authority already carried, so my upstream log was silently
   discarded. `a_real_effect_leaves_a_real_receipt` surfaced it. The redundant
   field is gone; `receipts()` delegates to the real log, with the override
   documented at the accessor.

**Still open, recorded in the delta:** the witness is an `Option`, so an
authority built outside the mediation layer spends silently — 98 construction
sites, 2 attach a log. And `ReceiptLog::append` cannot fail (in-memory `Vec`), so
"issuance fails if logging fails" is not yet expressible; it becomes real when
the log gains a durable sink.

brandon-coproduct:
### Added: FM-3 completed — no spend without a receipt

`Authority::spend` accepted a missing witness and recorded nothing. That is a
**fail-open audit**: the effect happens, the record does not, and every existing
test stays green because the effect still works. It now returns
`SpendError::Unwitnessed`.

Checked **before** the scope check, deliberately — if the scope check ran first,
a mis-scoped spend on an unwitnessed authority would report `ScopeMismatch` and
the absent log would stay invisible: the configuration defect masked by the
policy one.

#### I expected a 98-site refactor. It was six tests and one real defect.

Measuring first showed why. `PolicyEnforced::gate` calls `witnessed`, which
**replaces** whatever log an authority carried — so every construction site
feeding the mediation layer already gets the right log, and requiring one at
construction would have been ceremony. The only sites that matter are spends that
*bypass* `PolicyEnforced`, and a grep found exactly one in the workspace.

#### That one was a real hole

`Sandbox::spend_as` was an *associated function* — no `&self`, so no log in
scope, so nothing to witness with — and it spent directly. **Every sandbox write
authorised an effect and left no record.** Six write tests failing is what
surfaced it: the fail-closed spend converted a silent gap into a loud one, which
is the whole argument for making it fail closed.

`Sandbox` now owns a `ReceiptLog`. Both outcomes are recorded, not just
successes — `a_refused_sandbox_write_is_also_recorded` exists because a log
holding only successes answers "what was allowed" but not "what was attempted",
and the second question is the one an audit trail is for.

Adding the variant also broke an irrefutable `let SpendError::ScopeMismatch(why)`
in an existing test, which is the enum doing its job: a test asserting a refusal
now has to say *which* refusal.

**Limit, unchanged:** `ReceiptLog::append` cannot fail (in-memory `Vec`), so
"issuance fails if the log write fails" is still not expressible. It becomes real
when the log gains a durable sink.

*Also checked this turn:* a search result claimed Rust 1.95 shipped linear types
via `std::marker::MustMove`. `authority.rs` already records that claim as false;
re-verified against 1.96.1 — the trait does not exist. The affine design stands.

PR #2113, opened 2026-07-29
fix(firecracker): restore the Tier 3 token — omitting it deadlocked the launch

brandon-coproduct:
**#2107 broke pod launch on identity-enabled Firecracker nodes. This restores it.**
Found by running the real `nucleus-node` on the KVM box, which is the only way it
could have surfaced.

## What #2107 got wrong

It emitted `nucleus.sandbox_token` only for pods with no SVID, reasoning that an
identity-bearing pod proves itself from its SVID and never reads the token. That
is right about *what* such a pod needs and wrong about *when* it can get it:

```
wait_for_proxy_health(..)          <- guest must already be healthy
identity_registration(..)
WorkloadApiVsockBridge::start(..)  <- the SVID source, too late
```

With Tier 3 present the guest proved itself from the token, became healthy, and
fetched its SVID afterwards. Without it the guest has **no proof at the only
moment that matters**, exits fatally as PID 1, and the launch fails with
`proxy health check timed out`.

## Why no test caught it

The omission is **correct in isolation** and only wrong *in composition with the
spawn ordering*. Every unit test passed — including the ones I wrote for #2107,
which asserted exactly the behaviour that turned out to be fatal.

Observed end to end: the node built a cmdline carrying `workload_api_port` and no
`sandbox_token`, and the guest logged

```
failed to fetch identity: failed to connect to workload API: Connection reset by peer
FATAL: naked process detected — no sandbox proof available. Tried all 3 tiers.
```

After this change the same pod's guest proxy starts and stays up.

## The real fix, tracked separately

Start the bridge **before** the health check. The guest's identity is a
prerequisite for it being healthy, so producing it afterwards is the wrong order.
That is a restructure of `spawn_firecracker_pod`; when it lands, the conditional
comes back and three keys' worth of command-line exposure goes with it.

The ordering dependency is now written at the emission site, so the next person
to attempt this removal reads why it failed first. The test that asserted the old
behaviour asserts the new one **and says why**, rather than being deleted — the
claim it made was reasonable, and the reason it was wrong is the part worth
keeping.

## Verification

`cargo test -p nucleus-node` — 212 pass on Linux. Boot-verified through the real
node on Firecracker 1.16.1.


brandon-coproduct:
Superseded by #2114, which fixes the ordering itself rather than reverting the conditional: the workload API bridge now starts before the health check, so an identity-bearing pod can prove itself at the moment it must.

PR #2153, opened 2026-08-02
Bind the credential broker's socket at pod spawn

brandon-coproduct:
Every piece of the broker existed — framing with bounds, the accept loop, listener-bound identity, a `BrokerResponse` that **cannot** carry a credential because `Credential` has no `Serialize` impl — and nothing bound the socket. The source said so plainly:

> no accept loop binds this socket during pod spawn, so the guest cannot submit an envelope

This is that binding. `spawn_firecracker_pod` now runs the launch decisions the pure functions were written for: `decide_rollout` from the transport, `broker_identity` (returns `Option<PodIdentity>` so a caller cannot get a socket decision without the identity it speaks for), `prepare_socket`, then `serve_broker`. `on_start_failure` decides the failure mode — abort when credentials were withheld, because a guest denied its credentials with no way to ask cannot do its job.

## This is the host half only

The guest still cannot submit an envelope. Linux `-D warnings` now says exactly that: `serve_broker` and `serve_connection` are no longer dead, and the three items still unused are all the client half, including `request_over_socket`.

The compiler is stating *"the host serves, the guest cannot yet ask"* — which is the honest description of this change, and the guest client is the next piece.

## Two compile errors macOS could not see

Everything past `spawn_firecracker_pod`'s `#[cfg(target_os = "linux")]` guard isn't compiled on macOS, so a local check says nothing about it. Type-checked in an OrbStack Linux VM, which found `BrokerTransport::PerPodVsock` (the variant is `Vsock`) and `ApiError::Spec` (nucleus-node's is `Driver`; `Spec` is the tool-proxy's). **Both had already been reported as compiling clean.**

## The dead-code allow needs two layers

- **File-level, scoped to non-Linux**: on macOS the serving path is behind the cfg guard and legitimately unreachable, so warning is noise. On Linux it does not apply and the detector is live — if the spawn path stops binding the socket, `serve_broker` goes dead and CI says so.
- **Per-item on the three client-half functions**: dead on both platforms until the guest client exists, each naming what will use it.

Blanket file-level would hide the first; removing it entirely breaks macOS.

## Verification

Linux clippy `-D warnings` clean, macOS clippy clean, 44 broker tests, fmt, line ratchet `--strict`, mediation, sealed-home, verify-strict, dep ceiling.

### DIFF
--- a/crates/nucleus-node/src/broker_transport.rs
+++ b/crates/nucleus-node/src/broker_transport.rs
@@ -24,11 +24,22 @@
 //! at the limit rather than after it. The two bounds are not redundant: one
 //! protects the parser, the other protects the reader that feeds it.
 
-// Not yet reachable: no accept loop binds this socket during pod spawn, so the
-// guest cannot submit an envelope and credentials still arrive the old way.
-// CI denies warnings, and a bare dead_code warning would read as an oversight
-// rather than a stated gap. Every item is exercised by the tests below.
-#![cfg_attr(not(test), allow(dead_code))]
+// The SERVING path is live: `spawn_firecracker_pod` binds this socket and spawns
+// `serve_broker`. What is still unreached is the CLIENT half — the guest has no
+// way to submit an envelope yet — so the three items that await it carry their
+// own `allow(dead_code)` naming what will use them, rather than a file-level
+// allow that would also exempt the serving path.
+//
+// That distinction is the point. If the spawn path ever stops binding the
+// socket, `serve_broker` goes dead and `-D warnings` says so ON LINUX. A blanket
+// allow would have hidden exactly that.
+//
+// The file-level allow below is therefore conditioned on NOT being Linux:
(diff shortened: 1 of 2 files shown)


brandon-coproduct:
Closing: the socket was already bound. `start_broker_for_pod` exists at main.rs:1355 and is called at 2873, inside the same `spawn_firecracker_pod` this PR edited — so this added a SECOND listener on the same path.

The existing implementation is also better than what I wrote: it calls `check_enforcement_is_honest`, which is what turns `--broker-enforcing` into a refusal rather than a false claim on Firecracker, and it refuses to start when the pod's policy will not resolve rather than substituting a default lattice (my version used `unwrap_or_default()` — exactly the defect that comment warns about).

I trusted the stale comment in `broker_transport.rs` ("no accept loop binds this socket during pod spawn") instead of grepping for callers. The real gap is narrower and unchanged: the GUEST has no client, which is why `request_over_socket` is dead.

PR #2162, opened 2026-08-03
Two flagship proofs were never built, and a build failure reported success

brandon-coproduct:
The strongest-sounding evidence in this repo — machine-checked theorems with quoted axiom profiles — was the evidence least connected to anything that ran.

## Neither proof was compiled by any job

`CredentialNoninterferenceExtracted.lean` and `EgressConfinementExtracted.lean` appear in `aeneas-ifc-scoped.yml` **only in its `paths:` triggers** (lines 41–42, 55–56). The build step at 233–236 compiled four *other* libraries. Across every workflow there were **zero** `lake build` sites naming these two, or their generated dependencies `PortcullisCoreEgress` / `PortcullisCoreCredential` — which had no workflow mention at all.

So editing either proof fired a job that did not compile it.

They back `docs/production-delta.md` row 9 (egress confinement, 8 theorems) and row 36 — **FM-1 credential noninterference, the flagship formal claim**: *"Seven theorems … all `[propext, Classical.choice, Quot.sound]`, zero `sorry`, 0 opaque functions."* Those profiles were transcribed from some local run.

Everything else was already in place — both `lean_lib` targets declared, both generated directories committed, `#print axioms` written in both files. Only the build was missing.

## And the three that *were* built could not fail

Found while fixing the above. Every theorem target is piped to `tee` so the axiom audit can read the log, and GitHub's default `run` shell is `bash -e {0}` — `-e` **without** pipefail. The pipeline reports *tee's* exit status.

Demonstrated rather than argued:

```
bash -e           -c 'false | tee /dev/null; echo $?'   -> 0
bash -eo pipefail -c 'false | tee /dev/null; echo $?'   -> 1
```

None of the three Lean workflows sets `shell: bash`. So a type error in `IntegrityNoninterferenceExtracted`, `ConfidentialityNoninterferenceExtracted` or `MediationScopeExtracted` passed the build step — and then passed the axiom step too, because that step only asks whether a log *contains* `sorryAx`, and a log that was never written contains nothing.

## The axiom assertion was satisfied by silence

"The log does not contain `sorryAx`" is true of an empty file. Same shape as `check_art12_witnessed`, which asserts non-vacuity **first** for exactly this reason.

The step now requires positive evidence — each log must exist, be non-empty, and contain `#print axioms` output — before believing its silence. Verified against four log states rather than reasoned about:

| log | meaning | exit |
|---|---|---|
| `good.log` | real axiom output | 0 |
| `empty.log` | build never ran | 1 |
| `buildfail.log` | compile error | 1 |
| `dirty.log` | `sorryAx` present | 1 |

The middle two are the ones that used to pass.

## Deliberately not done

Pinning the exact axiom **set** against what the docs quote. That needs the real output from a run, and the documented profiles have never been produced by one — encoding them now would pin a transcription rather than a fact. The step prints the profiles prominently; pinning them is the next change, once there is ground truth to pin.

## Provenance

The missing builds were found by an adversarial subagent sweep and verified independently here (every `lake build` site enumerated; the two files' only mentions confirmed to be `paths:` triggers). The pipefail masking was found while implementing the fix.

The sweep's own verifier corrected two overreaches in the finding it confirmed, which is why it was worth trusting: a **literal** `sorry` token *is* caught by `portcullis-core-proven-lean.yml`'s whole-tree textual scan, so the gap was always the kernel check specifically.

### DIFF
--- a/.github/workflows/aeneas-ifc-scoped.yml
+++ b/.github/workflows/aeneas-ifc-scoped.yml
@@ -376,7 +376,29 @@ jobs:
           # (a real hole shows up as `sorryAx`); this is belt-and-suspenders and
           # must NOT false-positive on the file's own honest prose discussing
           # `sorryAx` / "sorry-free" (a plain substring grep does, and did).
-          if grep -rnE '(:=|=>|\bby\b|<;>|·|;|^)[[:space:]]*(sorry|admit)\b|\bnative_decide\b' \
+          # `native_decide` now shares the TACTIC-CONTEXT guard with sorry/admit.
+          # It did not, and the comment above already knew why that matters: this
+          # step fired on the two files' own prose EXPLAINING that native_decide
+          # is NOT used —
+          #
+          #   EgressConfinementExtracted.lean:268  `native_decide` ... is not used
+          #   CredentialNoninterferenceExtracted.lean:146  `native_decide` was
+          #     written first ... and then removed
+          #
+          # A gate that fires on the explanation of the property it checks is the
+          # same defect as one that never fires: neither is about the code.
+          #
+          # `^` was INSIDE the alternation group, where it does not anchor — so a
+          # bare tactic on its own line was missed entirely:
+          #
+          #     theorem t : True := by
+          #       sorry            <- NOT caught by the previous pattern
+          #
+          # That hole predates this change and applied to sorry/admit too. It is
+          # mitigated — a real hole shows up as `sorryAx` in the axiom audit above,
+          # which is the authoritative check — but a backstop that misses the most
+          # ordinary formatting of the thing it looks for is not backstopping.
+          if grep -rnE '((:=|=>|\bby\b|<;>|·|;)[[:space:]]*|^[[:space:]]*)(sorry|admit|native_decide)\b' \
                IntegrityNoninterferenceExtracted.lean \
                MediationScopeExtracted.lean \
                EgressConfinementExtracted.lean \
--- a/.github/workflows/aeneas-ifc-scoped.yml
+++ b/.github/workflows/aeneas-ifc-scoped.yml
@@ -321,7 +321,36 @@ jobs:
           # `|| true` because a grep matching nothing exits 1, and under pipefail
           # that would fail the assignment and RED a clean log — verified against
           # a clean log rather than assumed.
-          bad=$(grep -hoE "depends on axioms: \[[^]]*\]" $LOGS \
+          # FLATTEN NEWLINES FIRST. Lean WRAPS long axiom lists:
+          #
+          #   'thm' depends on axioms: [propext,
+          #    Classical.choice,
+          #    Quot.sound]
+          #
+          # A line-based `grep -o` never matches those — no closing `]` on the
+          # line — so it silently audits only the unwrapped ones. Measured on the
+          # real output: 3 of 7 credential theorems matched before this fix, and
+          # the step still reported success, because "found no bad axiom" is true
+          # of the four it never looked at.
+          #
+          # Every fixture I invented for this check was single-line. The real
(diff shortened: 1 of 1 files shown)


brandon-coproduct:
**Amended (3e9c11ab): the axiom check now also refuses anything outside the clean set.**

I originally declined to pin the axiom profile, reasoning that the documented values were never produced by a run — so encoding them would pin a transcription, CI would red for the wrong reason if reality differed, and the natural response to that is to edit the gate rather than read it.

That reasoning is right about **equality** and I over-applied it to **containment**. An allowlist needs no ground truth: it passes whichever subset is actually printed, cannot red for the wrong reason, and was available at the same cost.

**What it catches that the first version didn't:** `Lean.ofReduceBool` — introduced by `native_decide` — matches neither `sorryAx` nor `External`. The textual scan covers the four proof files and their generated dirs, so a `native_decide` in a dependency *outside* those paths would have landed in the profile unremarked. Also any custom `axiom` declaration.

`propext`, `Classical.choice`, `Quot.sound` are the standard clean-Lean set, so the allowlist isn't a guess about these particular files.

### A control caught a bug in the check itself

A `grep` matching nothing exits 1, and under `pipefail` that fails the assignment and reds a **clean** log. The first version did exactly that: all three failure cases behaved correctly, and both pass cases exited 1 with no output. Testing only the failures would have shipped a check that fails on everything — which then gets "fixed" by loosening it.

### Verified by running the real step

The step's shell body is extracted verbatim from the workflow and run against six log states, so what's tested is what CI executes:

| state | exit |
|---|---|
| clean profile | 0 |
| narrower subset | 0 |
| `native_decide` axiom | 1 |
| `sorryAx` | 1 |
| build failed | 1 |
| empty log | 1 |

Still deliberately **not** done: asserting the set *equals* what the docs quote. That needs ground truth from a run — and the profiles this step prints are exactly that ground truth.

brandon-coproduct:
**Ground truth obtained (a868970d) — and it falsified the check I had just written.**

I built both libraries locally, which is the run the previous commit said it was deferring for.

## The documented profiles are TRUE

All 7 `CredentialNoninterference` theorems and all 8 `EgressConfinement` theorems report exactly `[propext, Classical.choice, Quot.sound]`. Both libraries build clean. **The proofs were real; they were only ungated.**

## But my extraction was reading 3 of 7 theorems and reporting success

Lean *wraps* long axiom lists:

```
'thm' depends on axioms: [propext,
 Classical.choice,
 Quot.sound]
```

`grep -oE "depends on axioms: \[[^]]*\]"` matches line by line, and a wrapped list has no closing `]` on its line — so it matched none of them. Measured on real output: **3 of 7** credential theorems parsed, and the step still passed, because "found no axiom outside the allowlist" is true of the four it never looked at.

That is the same absence-satisfied-by-silence shape this PR exists to fix, one level down, introduced by me while fixing it. Every fixture I invented was single-line; I had never seen the real output.

Fixed by flattening newlines. **29 of 29 now parse.**

### The self-check that would have caught it instantly

Extracted count must equal printed count. A parser silently dropping entries now fails instead of passing — and it needs no magic number, since the log states both sides.

## Withdrawing my own follow-up

I said the next step was to pin the set by **equality** once a run produced it. Having the run: equality is the wrong check. A theorem that *drops* an axiom is a cleaner proof, and a gate that reds on an improvement teaches people to edit the gate. **Subset is the correct permanent form.** The equality follow-up is withdrawn rather than left as debt nobody should pay.

## Verified against the real logs

| input | exit |
|---|---|
| real wrapped logs (29 profiles) | 0 |
| single-line clean | 0 |
| wrapped list with `Lean.ofReduceBool` | 1 |
| empty log | 1 |

PR #2165, opened 2026-08-03
Every gate must fail on its own subject

brandon-coproduct:
Ten defects were found in this repo over two days and **every one was green**. The ones with the widest blast radius were not in the runtime — they were in the things watching it:

- two flagship Lean proofs no CI job compiled (#2162)
- three theorem builds whose failure reported success (`cmd | tee` under `bash -e`)
- the **vendor-neutrality gate** exiting 1 into a discarded pipeline (#2164)
- a cargo-mutants job printing *"All mutants caught by tests"* when the run had crashed

Each was found one at a time, by accident, while doing something else. **A gate that cannot fail is indistinguishable from a gate that passes**, and nothing was checking the difference.

For each gate: introduce a *real* violation of the property it names, assert non-zero, restore, assert zero. Both halves are required — a gate that fails on everything detects as little as one that fails on nothing, and only the restore half separates them.

## Its first run found two problems, both in itself

`check-line-ratchet.sh` reported as broken — it only exits 1 with `--strict`, and I probed it without the flag. `check-verify-strict.sh` likewise: the gate matches the dalek two-argument shape `.verify(msg, &sig)` and my probe wrote `.verify(m, s)`, which slips past it.

**Both gates were fine.** So the script now records how CI invokes each gate and **checks that against the workflow files** — probing a gate differently from CI tests something CI does not run, which is a recorded failure mode here and would otherwise have made this script its newest instance.

## Honest coverage, as a ratchet

Four gates probed, five not — and the five are **listed rather than omitted**, with their count as a ceiling that may only shrink. A meta-gate silently covering half the gates would be the exact vacuity it exists to find.

The uncovered five need cargo resolution or a manifest edit, which this script deliberately will not do — it must not leave a lockfile behind.

## Safety

- Refuses to run on a **dirty tree**: it edits real files and restores from a copy, and `git checkout -- <file>` has destroyed uncommitted work in this repo before.
- Restores under `trap … EXIT INT TERM` — a half-perturbed tree left by an interrupted run is worse than no check.
- Runs as its own CI job, not sharing a checkout with anything that writes into the tree.

## Verified

Perturbation: make `check-mediation.sh` always exit 0.

```
FAIL  check-mediation.sh — a raw Command::new on the agent path did NOT fail the gate (exit 0)
      The gate cannot detect the thing it is named for.
FAILED: 1 problem(s). A gate that cannot fail is not a gate.
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -114,6 +114,21 @@ jobs:
       - name: Forbid raw agent-path effect primitives outside the discharge-gated API
         run: scripts/check-mediation.sh
 
+  # Every gate above is a claim that something is checked. This checks the claim.
+  #
+  # It needs full history and a clean tree because it PERTURBS real files and
+  # restores them, so it runs as its own job rather than sharing a checkout with
+  # a job that might have written into the tree.
+  gates-can-fail:
+    name: Gates must fail on their own subject
+    runs-on: ubuntu-24.04
+    timeout-minutes: 10
+    steps:
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v6
+      - name: Every gate REDs on a real violation and GREENs when restored
+        shell: bash
+        run: scripts/check-gates-can-fail.sh
+
   ingest-hash-gate:
     name: No-unwitnessed-ingest gate (InputsAuthorized)
     runs-on: ubuntu-latest
--- a/scripts/check-gates-can-fail.sh
+++ b/scripts/check-gates-can-fail.sh
@@ -1,0 +1,226 @@
+#!/usr/bin/env bash
+# The gate of gates: every check-*.sh must FAIL on its own subject.
+#
+# WHY THIS EXISTS
+#
+# Over two days, ten defects in this repo were found and every one was green.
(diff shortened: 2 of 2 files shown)


brandon-coproduct:
**Coverage paid down 5 → 2 uncovered (035f051b).** Seven of nine gates now probed:

| gate | violation introduced |
|---|---|
| `line-ratchet` | 400 lines past the ceiling |
| `mediation` | a raw `Command::new` on the agent path |
| `sealed-home` | an un-allowlisted spawn in the sealed home |
| `verify-strict` | a non-strict dalek `.verify(msg, &sig)` |
| `failclosed-verifiers` | a `cfg(not(linux))` verifier returning `Ok(())` |
| `ingest-hashed` | an unwitnessed `.observe()` ingest |
| `sandbox-trusted-base` | a `pinned_by:` naming a nonexistent test |

All seven RED on the violation and GREEN when restored.

The three added here became tractable only once each gate's **detection was read rather than guessed at** — which is the same reason the first run misreported two gates as broken. Guessing at what a gate matches produces a probe that tests nothing; reading it takes a minute.

The remaining two (`dep-ceiling`, `wasm-closure`) need a `Cargo.toml` or lockfile edit, which this script deliberately will not do — it must not leave a lockfile behind.

PR #2166, opened 2026-08-03
A secret in the node's environment does not surface in the guest

brandon-coproduct:
**Increment 4a.** The node's environment holds every HMAC secret the runtime has; the guest is the thing the sandbox contains. Nothing checked that the first doesn't reach the second.

## The plumbing was never the blocker

I recorded this as needing a systemd unit override. It doesn't: `/etc/nucleus/node.env` already exists at mode 0600 and is `EnvironmentFile`'d into the unit by `provision::node_env_body`. CI appends one line.

That's the fourth time this session I called something a limit when it was a gap in what I'd looked at — after the `/dev/vsock` control, a stale memory about the discharge seal, and two claims about Rust's type system. All four were one grep away.

## The guest dumps; the host searches

The obvious design hands the workload the secret so it can `grep` for it — which puts the secret **in** the guest and destroys the property under test.

So the workload writes the enumerable places a node-held secret could surface — `/proc/cmdline`, its own environment, the proxy's environment, the pod spec baked into the rootfs — each behind a fence marker. The host, which knows the canary, greps that dump. A whole-filesystem grep can't work this way; those four are bounded and are where a leak would land.

## Three assertions, and the order is the point

The dump **ran** (fences present), it **captured** something (the decoy planted in the workload's environment), and only then that the canary is absent.

Without the first two, "the canary was not found" is satisfied by a workload that never started — the same absence-satisfied-by-silence shape as a cargo-mutants run reporting *"All mutants caught"* after crashing (#2164).

The check **bails rather than skips** when `NUCLEUS_E2E_CANARY` is unset. A silent skip is how this rots into a reassuring no-op.

## A hit names the section, never the value

A test that printed the secret into a CI log to prove the secret doesn't leak would be self-defeating — so a test asserts the failure message doesn't contain it.

## Testable without /dev/kvm

The decision is split into `assess_leak_dump(contents, canary, decoy)` so it runs anywhere. A booted pod needs KVM; without this split every claim about the check would rest on reading it.

| state | result |
|---|---|
| clean dump | ok |
| no fences (never ran) | refused |
| fences but no decoy (read nowhere) | refused |
| truncated mid-dump | refused |
| canary present | refused — value withheld, section named |

## Scope

Deliberately **not** a `credentialed_egress` entry: on top of #2161 that would trip the startup refusal unless tier2 also enables `--broker-listen`. This is 4a — *a node-held secret does not surface in the guest* — which stands on its own and covers the kernel-cmdline path where `nucleus.auth_secret` already rides. The credential-bearing round trip is 4b.

fmt, gates, macOS clippy `--workspace --all-targets --all-features -D warnings`, `cargo test --all-features` (222 suites, 0 failures).

### DIFF
--- a/.github/workflows/quickstart-boot.yml
+++ b/.github/workflows/quickstart-boot.yml
@@ -310,11 +310,31 @@ jobs:
           # release artifacts here would test a build nobody is proposing.
           nucleus setup --skip-verify --artifacts local
 
+      - name: Plant a canary in the NODE's environment
+        shell: bash
+        run: |
+          # The node's environment holds every HMAC secret the runtime has, and
+          # the guest is the thing the sandbox contains. This plants a value
+          # there so `verify --tier2` can assert it does NOT surface in the
+          # guest — an assertion that is meaningless without a real secret to
+          # look for, which is why the check REFUSES rather than skips when
+          # NUCLEUS_E2E_CANARY is unset.
+          #
+          # /etc/nucleus/node.env already exists at mode 0600 and is
+          # EnvironmentFile'd into the systemd unit — see provision::node_env_body.
+          CANARY="nucleus-e2e-canary-$(openssl rand -hex 12)"
+          echo "NUCLEUS_E2E_CANARY=$CANARY" | sudo tee -a /etc/nucleus/node.env >/dev/null
+          sudo systemctl restart nucleus-node || sudo nucleus start || true
+          # Exported for the verifier, NOT written anywhere the guest can read.
+          echo "NUCLEUS_E2E_CANARY=$CANARY" >> "$GITHUB_ENV"
+
       - name: Boot a real pod and assert what the guest did
+        shell: bash
         run: |
           # The whole point. Asserts, in order: the pod is created; a permitted
-          # command runs INSIDE the microVM; a forbidden command is DENIED; and
-          # the guest fetched its SVID and its task token over vsock.
+          # command runs INSIDE the microVM; a forbidden command is DENIED; the
+          # guest fetched its SVID and its task token over vsock; and no secret
+          # from the node's environment surfaced inside the guest.
           nucleus verify --tier2 --here
 
       - name: The in-guest posture probe actually ran (presence, not just verdict)
--- a/crates/nucleus-cli/src/verify.rs
+++ b/crates/nucleus-cli/src/verify.rs
@@ -380,41 +380,6 @@ struct CreatePodResponse {
 /// The DLC-D labels provision verified admission for THIS pod only (the node
 /// forwards them to the pod's tool-proxy as `NUCLEUS_DLC_*`); the issuer key
 /// doubles as its own trust anchor, matching dlc-d's principal convention.
-/// A value planted where the guest CAN legitimately see it.
-///
-/// The positive control for [`check_credential_absent_from_guest`]. Every
-/// assertion there is that the canary is ABSENT, and a dump that captured
-/// nothing satisfies all of them — so the sweep must first be shown to find
-/// something it is supposed to find.
-const DECOY: &str = "nucleus-e2e-decoy-6f21a0";
-
-/// What the pod's workload prints, and why it prints rather than searches.
-///
-/// # The guest must never learn the canary
-///
-/// The obvious design hands the workload the secret so it can `grep` for it —
-/// which puts the secret IN the guest and destroys the property the test exists
-/// to check. So the guest DUMPS and the host SEARCHES: the workload writes the
-/// enumerable places a node-held secret could surface, and
-/// `check_credential_absent_from_guest`, which knows the canary, greps that.
-///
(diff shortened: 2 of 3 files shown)


brandon-coproduct:
**Converting to draft: the design is wrong, and CI caught it.**

```
Error: the guest log has no E2E-DUMP-BEGIN marker, so the leak-site dump did
not run to completion. 'the canary was not found' would mean 'nobody looked'.
```

The canary was set correctly (`NUCLEUS_E2E_CANARY: nucleus-e2e-canary-23ed…`), the node picked it up, and the pod booted — `check_guest_facts` passed before this. The workload simply never ran.

## Why

`scripts/firecracker/build-rootfs.sh:304` — `cp "$POD_SPEC" "$ROOTFS_DIR/etc/nucleus/pod.yaml"`. On Firecracker the pod spec is **baked into the rootfs at image build time**, and the guest's tool-proxy reads `/etc/nucleus/pod.yaml`. The `workload` field I added to `create_pod`'s API request never reaches the guest at all.

**I already knew this.** It's the same fact I cited in #2158 to argue that `store_from_node_environment` must not read `spec.spec.credentials.env` — because a credential taken from the pod spec would ship inside the image. I wrote that into a doc comment and then built this increment on its opposite.

## The control is the only reason this isn't a false green

Without the "did the dump actually run" assertion, this would have reported **"no canary found in the guest — pass"** while nothing was ever inspected. The check would have been green, permanent, and meaningless.

That is precisely the failure class this PR exists to close, and it very nearly shipped as its own instance.

## Options, none of which is a patch

1. **Bake a dump workload into a tier2-specific rootfs.** Faithful, but puts test scaffolding in an image build.
2. **Have `guest-init` emit the leak-site dump**, conditionally. It already writes to the console — that's how `spiffe://` reaches `firecracker.log`. Cost: test-only code on the production guest path.
3. **Scope 4a to what the host controls** — assert the canary is absent from the kernel cmdline the node constructs, from any guest spec it writes, and from the rootfs. Checkable host-side with no workload, and still a real property, but it tests what the host *emits* rather than what the guest can *see*.

(3) is tractable now; (2) is the fuller check. Either is a redesign, so this PR should not merge in its current shape.

brandon-coproduct:
Triaged + rebased onto main (compiles, no conflicts) and ran the real boot. **Diagnosis: not a leak — a stale harness wiring.**

The boot passes every stage (pod create, SVID, allowed/denied ops, verified-admission) and then the credential-absence sweep correctly fails its own **non-vacuity guard**:
```
Error: the guest log has no E2E-DUMP-BEGIN marker, so the leak-site dump did not
run to completion. 'the canary was not found' would mean 'nobody looked'.
```
So the guest's `LEAK_SITE_DUMP` output isn't reaching the log `check_credential_absent_from_guest` reads (the `find -newermt … | head -1` guest-log heuristic + the workload-exec/serial-capture path have evolved since this was opened on 2026-08-03). The property and the decoy positive-control design are sound; the wiring that gets the guest dump into the verifier's log needs re-fitting to the current path before this can go green.

Keeping as **draft**. Whoever resumes: confirm how the workload's stdout currently lands in the located guest log, then point the dump/marker capture at that path.

brandon-coproduct:
## Root cause found (deeper than a log path)

The firecracker.log from the failing boot shows the workload that actually ran is **`nucleus-workload-probe`** (`[workload] NUCLEUS_WORKLOAD_PROBE: PASS`, egress checks), **not** this PR's `/bin/sh` leak-site dump. A pod runs exactly one workload, and the CI boot **bakes the probe as the workload** via the rootfs `/etc/nucleus/pod.yaml` (`build-rootfs.sh:317`, from `examples/openclaw-demo/firecracker-pod.yaml`). So the `workload.command` this PR sets in its `POST /v1/pods` spec is overridden — the dump never executes, hence the non-vacuity guard correctly fires (`no E2E-DUMP-BEGIN marker`).

**This is a structural gap in the approach**, not a small bug: dump-as-a-separate-workload cannot coexist with the baked probe workload.

## Fix options (both multi-component; need boot-verify iteration)

1. **Fold the sweep into `nucleus-workload-probe`, gated.** The probe already runs and streams to the console. Add a leak-site emit gated behind an env flag (e.g. `NUCLEUS_E2E_DUMP`) so it only fires in the e2e boot — otherwise dumping `/proc/1/environ` values to the console every boot would itself put the proxy's secret on the console (the very property this guards). Requires: probe change + a test pod-spec/env that sets the flag + decoy + `verify.rs` greps the probe's `[workload]` output.

2. **Guest searches for the canary *prefix pattern* (`nucleus-e2e-canary-`), reports ABSENT/PRESENT only.** Keeps secret *values* off the console and keeps the full canary out of the guest (only the non-secret prefix is known guest-side); probe emits a boolean + a decoy-pattern hit for non-vacuity. A cleaner design but a departure from the host-searches-dump model.

Recommend option 1 (least surprising). Keeping as draft.

PR #2192, opened 2026-08-06
Machine v2's oracle mirrors had no faithfulness theorem — prove all five

brandon-coproduct:
## The gap

`IdentityOracleMirror` anchored **one** oracle (the v1 guard). Machine v2 grew a floating label, and its mirrors — the confidentiality lattice, its join, its flows-to, the material-label table, the principal ceiling — were transcribed into the iris tree with **nothing** proving they match the shipped extraction.

That's the sharpest form of the objection: one wrong entry in `matLabel` makes the v2 noninterference theorem a *true statement about the wrong system*, with every gate green. A proof that's real and a claim that's worthless is worse than having neither.

## The fix

Five faithfulness theorems against the Aeneas-extracted functions that ship — `cjoin_faithful`, `cflows_faithful`, `mat_label_faithful` (all 11 materials), `ceiling_faithful`, `may_deliver_faithful` (the full 11×3 product) — each by exhaustive case analysis, so a wrong entry **fails to compile**. Perturbation-verified: flipping `ProxyAuthSecret` Internal→Public breaks `mat_label_faithful`'s `rfl`.

`reaches_is_a_flow_faithful` additionally proves against the extraction what the iris tree only asserted in a docstring: the v1 delivery guard **is** the v2 flow decision at the workload's ceiling.

## Deliberate omission

No faithfulness theorem for the rank, with the reason in the file: extracted `crank` returns an Aeneas `Std.U8` while a portable mirror must return `Nat`, so stating equality threads a scalar conversion that adds TCB without adding assurance. The rank is internal and fully pinned by the two composites, which are exhaustive over all argument pairs.

## Scope

This anchors the mirror **definitions**. That the *copies* in `capability-primitive/iris` are these definitions cannot be proved here — a Lean proof can't see another repository — and is gated there by a companion parity check.

Added to the proven-tier build list so it can't rot.

### DIFF
--- a/.github/workflows/portcullis-core-proven-lean.yml
+++ b/.github/workflows/portcullis-core-proven-lean.yml
@@ -108,7 +108,8 @@ jobs:
             PodMachineSpike \
             IdentityOracleMirror \
             ConfidentialityOracleMirror \
-            IntegrityOracleMirror
+            IntegrityOracleMirror \
+            ChannelOracleMirror
           # CapabilityResiduatedQuantaleProofs: the residuation adjunction
           # (a⊗b≤c ⟺ b≤a⊸c) for capability, proven over the Aeneas-extracted
           # capability_quantale slice — the enriching value V of the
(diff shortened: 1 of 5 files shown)


brandon-coproduct:
Superseded: all five oracle-mirror faithfulness theorems (Identity/Confidentiality/Integrity/Channel + PodCrossView) are already in main — the .lean files exist with their theorems (5/14/8/5) and are in the proven-lean build list. This PR's contribution has landed via later work; closing as redundant.

PR #2194, opened 2026-08-06
Pod self-identity, and ManagePods scoped to descendants

brandon-coproduct:
Two linked defects from a cross-pod audit. The second cannot be fixed without the first.

## 1. `NUCLEUS_POD_ID` had a consumer and no producer

`lockdown_client::apply_scope` reads it to decide whether a `pod:<id>`-scoped lockdown is aimed at this pod. **Nothing in the tree ever set it** — repo-wide, one occurrence: the read.

So the `Some(my_id) == target` branch was unreachable and `apply_scope` always fell to `None => true` (*"conservatively lock"*). **A lockdown aimed at ONE pod locked down EVERY pod on the node.** Fail-safe, which is why it survived, and silently useless as targeting: an operator scoping an incident to one tenant did not.

guest-init now derives it from the SVID it already fetches — **from the transport, not an env var**: the SVID arrives over the per-pod vsock socket the host creates per VM, so the identifier is not something the guest can restate. Only the workload-API form (`ns/pods/sa/{uuid}`) parses; the *other* SPIFFE ID a pod has (`ns/{namespace}/sa/{metadata.name}`) is rejected, because two pods can share a `metadata.name` and scoping would match the wrong set.

## 2. `ManagePods` was node-wide by accident

`check_manage_pods` asks whether the caller *holds* the capability, never *which pods* it authorises. A pod with it could list every tenant's pods, read any pod's logs, and cancel any pod — and `list_sub_pods` returning the whole node list is how a caller *learned* the UUIDs to pass.

The node had the answer all along: `parent_pod_id` has always been tracked and was consulted only for cascade-cancel. `PodInfo` now carries it, and the three handlers enforce descendant-only management. `get_pod_logs`/`cancel_sub_pod` re-resolve the target through the node's registry so parentage never comes from the request, and *"not yours"* and *"doesn't exist"* return the **same** refusal so UUIDs can't be enumerated.

## Where this is enforced, and the residual

**Proxy-side**, where `check_manage_pods` already lives: the architecture treats the agent as untrusted and the proxy as the guest's enforcement point, and the agent is who this stops. A **compromised proxy** can bypass it, because every proxy holds the node-wide auth secret. Making the *node* enforce it needs per-caller identity at the node API, which doesn't exist today — every pod signs as the literal actor `"tool-proxy"` with the same shared secret (`node_client.rs:89`). That's the next arc.

## The named gap

With `NUCLEUS_POD_ID` unset the check can't be made and behaviour is unchanged rather than fail-closed — otherwise this breaks pod management wherever identity was never fetched (local mode). Logged loudly, and **pinned by a test** whose failure would signal someone tightened it deliberately. Part 1 makes the real Firecracker path have identity.

## Tests (10)

Four ownership assertions, one pinning the named gap, four on the identity parser (including rejecting the namespace form), and **two structural** tests that read the source and fail if the wiring is dropped — the defect in part 1 was exactly a consumer with no producer.

Verified to fail for the right reason: neutering the ownership predicate reds exactly the two cross-tenant assertions. Full tool-proxy suite 299 passed; Linux `cargo check` on guest-init (a Linux-only target) passes.

### DIFF
--- a/crates/nucleus-guest-init/src/identity.rs
+++ b/crates/nucleus-guest-init/src/identity.rs
@@ -209,6 +209,41 @@ pub fn fetch_task_token(port: u32) -> Result<TaskTokenResponse, String> {
 /// and writes the certificate and key to the identity directory.
 ///
 /// Returns the SPIFFE ID on success.
+/// The pod's own UUID, read out of the SPIFFE ID the host just served it.
+///
+/// # Why this exists, and why it is not an environment variable from the host
+///
+/// The tool-proxy reads `NUCLEUS_POD_ID` to decide whether a pod-scoped lockdown
+/// command is aimed at it (`lockdown_client::apply_scope`). Nothing in the tree
+/// ever set that variable — so the `Some(my_id) => my_id == target_pod` branch was
+/// unreachable, and `apply_scope` fell to its `None => true` case: a lockdown
+/// aimed at ONE pod locked down EVERY pod on the node. Fail-safe, and silently
+/// non-functional as a targeting mechanism.
+///
+/// The identity is taken from the SVID rather than injected by the host as an
+/// env var because the SVID arrives over the per-pod vsock socket that the host
+/// creates per VM — the guest cannot forge which socket it is connected to,
+/// whereas any value merely placed in the environment is only as good as
+/// everything that can write to the environment. This is the same reasoning that
+/// makes the workload API's `pod_id` trustworthy host-side: the identifier comes
+/// from the transport, not from the payload.
+///
+/// Returns `None` rather than guessing when the ID is not the expected shape —
+/// a wrong pod id is worse than no pod id, because `apply_scope`'s `None` branch
+/// is the safe one.
+#[must_use]
+pub fn pod_id_from_spiffe(spiffe_id: &str) -> Option<String> {
+    // `spiffe://{trust_domain}/ns/pods/sa/{pod_id}` — see
+    // nucleus_node::workload_api_vsock, which mints exactly this form.
+    let rest = spiffe_id.strip_prefix("spiffe://")?;
+    let (_trust_domain, path) = rest.split_once('/')?;
+    let id = path.strip_prefix("ns/pods/sa/")?;
+    if id.is_empty() || id.contains('/') {
+        return None;
+    }
+    Some(id.to_string())
+}
+
 pub fn fetch_identity(port: u32) -> Result<String, String> {
     // Create identity directory
     fs::create_dir_all(IDENTITY_DIR)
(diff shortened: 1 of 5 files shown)


brandon-coproduct:
**Converting to draft — this PR breaks sub-pod management as it stands.** Found
while building the node-side enforcement that was meant to follow it. My own
defect; recording the evidence rather than quietly patching it.

## The problem

`may_manage` filters on `parent_pod_id`:

```rust
pod.parent_pod_id.map(|p| p.to_string()).as_deref() == Some(me) || pod.id.to_string() == me
```

But **nothing in this repository ever sets that field.** Both node entry points
derive it from a client-supplied header that no in-repo client sends:

- `crates/nucleus-node/src/main.rs:995` — HTTP: `headers.get("x-nucleus-parent-pod-id")`
- `crates/nucleus-node/src/main.rs:3376` — gRPC: the same key from request metadata

The proxy's `create_sub_pod` → `node.create_pod(..)` → `post_json("/v1/pods", ..)`
attaches signing headers and nothing else. A repo-wide grep finds only those two
*readers* and no writer.

So `parent_pod_id` is `None` for every pod created through the tool-proxy, and
`pod.id == me` is false when the caller is a parent listing its children. The
filter is therefore **false for every pod**.

## Why the tests did not catch it

This PR is also what makes `NUCLEUS_POD_ID` get set, in guest-init. The unit
tests never run guest-init, so `me` is `None` in tests and `may_manage` takes its
early-return-`true` path — the branch that preserves old behaviour. Every test
exercises the case where the check is disabled. On a real Firecracker deployment
`NUCLEUS_POD_ID` *is* set, `me` is `Some`, and `list_sub_pods` returns an empty
list while `get_pod_logs` / `cancel_sub_pod` refuse everything.

A gate that is green because the field it reads never moves — the same shape as
`a-gate-that-hardcodes-a-dimension`.

## Related pre-existing bug

Cascade-cancel at `main.rs:3084` also reads `pod.parent_pod_id`. On the same
evidence it has never had a non-`None` value to work with, so cascading
almost certainly never cascaded. That predates this PR.

## The fix, and why it belongs elsewhere

Populating the header from the proxy would be the small fix, and it would be the
wrong one: `x-nucleus-parent-pod-id` is **unauthenticated**, so lineage would be
client-declared and any pod could plant children under a victim or disown its
own. Building an ownership check on a forgeable field is worse than having no
check, because it reads as enforcement.

The right fix is for the node to derive `parent_pod_id` from the **identified
caller** — which is exactly what #2196 makes possible. I am implementing that
now, together with moving the ownership check node-side (this PR's documented
residual). That work supersedes the proxy-side filter here.

This PR stays draft until then. Its other half — pod self-identity in guest-init
— is still good and is a dependency of the replacement.

brandon-coproduct:
Closing as **superseded by #2199**, not abandoned — the good half of this PR
lives on there.

## Why this one cannot land as written

The proxy-side ownership filter here reads `parent_pod_id`, and nothing in the
repository ever *set* that field: both node entry points derived it from
`x-nucleus-parent-pod-id`, a header no in-repo client sends. So the filter was
false for every pod. Worse, this PR is also what causes `NUCLEUS_POD_ID` to be
set, so it would have switched the check on and simultaneously guaranteed it
matched nothing — an empty pod list and refused log reads on any real
deployment. The tests passed because they never run guest-init, so `me` was
always `None` and every test took the early-return path where the check is
disabled.

## Why the small fix would have been the wrong one

Sending the header from the proxy would have made the filter *work* while
leaving lineage **client-declared**. `x-nucleus-parent-pod-id` is
unauthenticated, so any caller could name any parent — planting children under a
victim or disowning its own. An ownership check reading a forgeable field is
worse than no check, because it reads as enforcement.

## What replaced it

- **#2199** — the node derives `parent_pod_id` from the *identified* caller and
  enforces ownership node-side, which also retires the residual this PR
  documented: every proxy holds the node-wide auth secret, so a proxy-side check
  does not survive a compromised proxy.
- **Pod self-identity in guest-init**, the genuinely good half of this PR, was
  carried forward as a dependency of that work.

It is also now a theorem rather than a judgement call:
`CrossPodShared.client_declared_lineage_breaks_ni` in the capability-primitive
corpus proves that a pod able to declare its own parent breaks noninterference,
and `CrossPodHost.no_pod_can_forge_lineage` derives that the node's check
prevents it.

PR #2219, opened 2026-08-08
ci: sccache compiler-cache prototype (experiment, non-required)

brandon-coproduct:
Prototype for evaluating **compiler-level** caching (sccache) on top of the dependency caching we already have — the optional item from the CI audit.

## What it measures

Our required jobs cache **deps** (Swatinem/rust-cache via `setup-rust-toolchain`): a run whose `Cargo.lock` is unchanged skips recompiling *dependencies*, but still rebuilds **this workspace's own crates** from scratch. sccache is compiler-level — it content-addresses each `rustc` invocation and reuses the object when inputs match, so it can also skip recompiling unchanged **first-party** crates across runs and concurrent jobs. This job runs a representative heavy `cargo check` under sccache and prints `sccache --show-stats` so we can see the **warm hit rate** and wall-clock, then decide whether to fold it into Clippy/Tests.

## How to read it

- **First run is cold** (~0% hits — it populates the sccache GHA cache). This PR triggers one.
- **Re-run via `workflow_dispatch`** (or after merge to main) to read the **warm** hit rate — that's the number that decides adoption. Both the compile wall-clock and the sccache stats land in the job's **step summary**.

## The coexistence config (what we'd actually adopt)

rust-cache and sccache overlap on the target dir, and the repo cache is quota-bounded, so I split responsibilities:
- rust-cache keeps **only** the registry/index (`cache-targets: false`), with a **distinct `prefix-key`** so this experiment never shares or evicts the required jobs' caches.
- sccache owns the compilation-object cache.
- `CARGO_INCREMENTAL: 0` — **mandatory**: sccache doesn't cache incrementally-compiled crates, so without it it's silently bypassed for exactly the first-party crates this targets (incremental is pointless in ephemeral CI anyway).

## Safety

Separate workflow, **never in the required set**, cannot block the merge queue. Triggers only on its own file (this PR), merges to main (to seed the cache), and manual dispatch — **not** on every code PR, so it doesn't burn CI minutes as a standing cost.

## Decision after measuring

If the warm hit rate and wall-clock win is real (the literature reports up to ~50% over rust-cache alone at high hit rates), fold this config into the required `Clippy`/`Tests`/`Doctests` jobs. If not, delete this one file. Either way the experiment is free of risk to the required set.

### DIFF
--- a/.github/workflows/sccache-prototype.yml
+++ b/.github/workflows/sccache-prototype.yml
@@ -1,0 +1,125 @@
+# sccache compiler-cache PROTOTYPE — an experiment, deliberately NOT required.
+#
+# WHAT THIS MEASURES
+#
+# Our required jobs already cache dependencies via Swatinem/rust-cache (through
+# `setup-rust-toolchain`'s `cache: true`). That caches the cargo registry/index
+# and the target dir keyed on Cargo.lock — so a run whose deps are unchanged
+# skips recompiling THEM, but still recompiles this workspace's OWN crates from
+# scratch. sccache is compiler-level: it content-addresses each rustc invocation
+# and reuses the object when the inputs match, so it can also skip recompiling
+# unchanged first-party crates across runs and across concurrent jobs.
+#
+# This job runs a representative heavy compile with sccache and prints
+# `sccache --show-stats`, so the hit rate and wall-clock can be compared against
+# the required `Clippy`/`Tests` jobs BEFORE deciding whether to fold sccache
+# into them. It is safe: separate workflow, never in the required set, cannot
+# block the merge queue.
+#
+# HOW TO READ THE RESULT
+#
+# The FIRST run is cold — it populates the sccache GHA cache, so expect ~0%
+# hits. Re-run it (push again, or `workflow_dispatch`) to see the WARM hit rate,
+# which is the number that matters. The step summary reports both compile
+# wall-clock and the sccache stats.
+#
+# THE COEXISTENCE CHOICE (the config we would actually adopt)
+#
+# rust-cache and sccache overlap on the target dir, and the repo cache is quota-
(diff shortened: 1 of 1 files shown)


brandon-coproduct:
## Measured — and the verdict is: **not worth adopting with the GHA backend.**

Two runs, identical crate sources, same sccache GHA cache scope:

| | Cold (run 1) | Warm (run 2) |
|---|---|---|
| Compile wall-clock | **168 s** | **153 s** |
| Cache hit rate (overall) | 0 % | **53.5 %** |
| Cache hit rate (Rust) | 0 % | 66.3 % |
| Cache hit rate (C/C++) | 0 % | 29.3 % |
| Objects hit / total | 0 / 904 | 484 / 904 |

**The hit rate is real (53%), but it converts to only ~9% wall-clock (168→153s).** The reason is the documented sccache+GHA caveat: the GitHub Actions cache backend fetches each of the 484 cached objects over the network, and that transfer latency eats most of the compute the cache saved. The 420 misses (heaviest first-party crates + build-script/proc-macro non-cacheables) still compile fully.

Worse, this understates the gap vs. what we already run: the required jobs use `rust-cache` with **target-dir caching ON**, so on the common case (deps unchanged) they restore prebuilt dependencies in **one** bulk fetch and recompile only first-party crates — likely faster than sccache's hundreds-of-small-fetches over the same slow backend.

**Conclusion:** a 9% win doesn't justify the added moving parts, the 10 GB quota competition, and the `CARGO_INCREMENTAL=0` constraint. sccache would only pay off here with a **faster cache backend** (S3 / Depot / self-hosted), which is infrastructure we don't have. Keeping the existing `rust-cache`.

Closing this experiment; it answered its question. Reopen (or ping) if we ever add an S3-backed cache and want to re-measure — the config here is the starting point.

PR #2220, opened 2026-08-08
fix: wire three declared-but-dead config surfaces (audit findings #1-#3)

brandon-coproduct:
The skeptical audit's top cluster: three places where a declared surface *reads as wired* via its doc-comment / `#[allow]` annotation but had **no consuming call site**. Each is now wired, with a test proving the knob bites.

## 1. `register_executor_pubkey()` was never called — `nucleus-node/trust_gate.rs`
Its `#[allow(dead_code)]` comment falsely said *"Called at startup"*. Every receipt POST is signed with the per-executor Ed25519 key (`X-Nucleus-Executor-Sig`) but ships **no inline pubkey**, so the trust-service could never verify those signatures — it never received the enrollment. **Fix:** called once from `main()` before serving, gated on `is_enabled()`; false `#[allow]` removed. (HMAC auth was unaffected — this was an inert *additive* identity layer, not an auth break.)

## 2. `NetworkSpec.mime_allow` — parsed but never read — `nucleus-spec` / `nucleus-tool-proxy`
The per-pod web_fetch MIME allowlist was **silently ignored**; the built-in `ALLOWED_MIME_PREFIXES` always applied despite a doc-comment promising the override. **Fix:** `check_mime_type` now takes an `Option<&[String]>` override (`Some` **replaces** the built-in list); both call sites (HTTP + MCP) pass `state.web_fetch_mime_allow`.

## 3. `NetworkSpec.max_response_bytes` — parsed but never read
The body cap was always the CLI default, and the doc falsely claimed a **10 MiB** default (real: **5 MiB**). **Fix:** effective cap is now the per-pod value or the CLI default; doc corrected.

## Structure
All web_fetch config resolution (dns/url/mime/max) is consolidated into `web_fetch_policy::resolve_web_fetch_config` so the construction site threads one value, not four — which also keeps `tool-proxy/main.rs` under its line ceiling.

## Tests (the knobs bite)
- `a_pod_mime_override_replaces_the_builtin_list` — an override of `["image/"]` ADMITS `image/png` (built-in blocks) and REJECTS `text/html` (built-in allows).
- `an_empty_mime_override_denies_everything_typed`.
- `a_pod_response_cap_overrides_the_configured_default`.

## Verified
Linux `clippy -D warnings` clean + tests green over `nucleus-node`, `nucleus-tool-proxy`, `nucleus-spec`; `fmt --all --check` + line-ratchet pass.

## Not in this PR
The other 4 audit findings (all Low): Python SDK `acknowledge()` false audit claim, lockdown `affected_pods` count placeholder, `shutdown_otel()` never called, quickstart-boot path-filter gap. Tracked for a follow-up.

### DIFF
--- a/crates/nucleus-node/src/main.rs
+++ b/crates/nucleus-node/src/main.rs
@@ -751,6 +751,13 @@ async fn main() -> Result<(), ApiError> {
         lockdown_tx: tokio::sync::broadcast::channel::<proto::LockdownCommand>(16).0,
     };
 
+    // Enroll this executor's Ed25519 public key with the trust-service, once,
+    // before serving. Every receipt POST is signed with the matching private
+    // key (`X-Nucleus-Executor-Sig`) but ships no inline pubkey, so the
+    // trust-service can only verify those signatures if it learned the key from
+    // this enrollment first. A no-op when the trust gate is disabled.
+    trust_gate::register_executor_pubkey(&state.trust_gate, &state.http_client).await;
+
     // Routes that require HMAC auth
     let authenticated_routes = Router::new()
         .route("/v1/pods", post(create_pod).get(pod_api::list_pods))
--- a/crates/nucleus-node/src/trust_gate.rs
+++ b/crates/nucleus-node/src/trust_gate.rs
@@ -1180,10 +1180,11 @@ fn hmac_sha256_hex(secret: &[u8], data: &[u8]) -> String {
 
 /// Register this executor's Ed25519 public key with the trust-service.
 ///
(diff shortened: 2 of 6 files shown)


brandon-coproduct:
Superseded by #2224, which lands this fix together with the other web_fetch hardening PR. They refactored the same tool-proxy startup block, so they mutually conflicted in the merge queue and each had gone stale past the line-ratchet ceiling; #2224 resolves both into one shorter, ratchet-passing change. Keeping this open until #2224 merges.

brandon-coproduct:
Landed via #2224 (merged). Closing.

PR #2227, opened 2026-08-10
north-star Phase 0: C5 value-commitment (increment 1) + C5 disconnect diagnosis + C4 live-egress correction

brandon-coproduct:
## Phase 0 of the declassification-unification program (doc/ledger honesty; no live-egress behavior change)

Full plan: the C4/C5/C7 declassification-unification program. This is **Phase 0 only** — record the findings and correct the overstatement. **No egress semantics change.** Boot/live-egress changes are Phases 2+.

### What lands here
1. **C5 increment 1 (already committed, rebased onto current main).** `DeclassificationToken` now carries a signed `content_commitment: [u8;32]` (SHA-256 of the exact authorized bytes) bound into `canonical_bytes` (v2→v3; v2 signatures no longer verify). A governor signature now authorizes exactly ONE value. This closes the *signing-layer* half of the earlier false C5 justification. Apply-time enforcement + the four-run theorem remain blocked on the two-graph disconnect (Phase 1–5).
2. **C5 disconnect diagnosis note.** Records that value-binding is blocked by an *architectural* two-graph split (token → kernel `flow_graph`, which the tool-proxy never populates; live egress reads a separate `FlowTracker`), not a missing theorem.
3. **C4 correction (honesty fix).** The C4 note claimed "the path is LIVE: `POST /v1/declassify` applies governor-signed tokens." Verified false: the sole production wrapper of `apply_declassification_token` is the tool-proxy governor endpoint, and on the shipping tool-proxy it is **inert** — ingest populates a separate `FlowTracker`, not the kernel `flow_graph` the token resolves against, so apply returns `NodeNotFound` and no egress verdict changes. C4 **stays PROVED** (the single-use property is genuinely proved over the Aeneas-extracted core and enforced at the kernel API; inert-on-one-deployment does not *falsify* a single-use claim). The row now cites the kernel-API enforcement test (`kernel_token.rs`) instead of the inert wrapper, and the note discloses the inertness and **retracts** the false sentence on the record.

### Fork decided (async, for your veto)
**C4: wording-precision fix, keep PROVED** — runner-up was demote PROVED→NOT-YET (would have raised the ratchet 5→6). Rejected: a demotion would understate a genuinely-proved-and-kernel-API-enforced single-use property and conflate "inert on the tool-proxy deployment" with "unproved" — as dishonest as inflating a false row. Fully reversible (doc-only): if you judge the tool-proxy egress path is the only "live path" that counts for a flagship row, flip to demotion + ratchet 5→6 in one commit.

### Gates (all green locally)
- `check-north-star-ledger.sh`: OK — 9 clauses, 5 NOT-YET (unchanged; no ratchet move; C1 stays PROVED from current main)
- `check-declassify-sink-scope-enforced.sh`: OK (22 declassify tests pass; sink scope enforced)
- `check-declassify-governor-keys-sealed.sh`: OK
- `cargo fmt/clippy -p portcullis-core`: clean (this PR also fixes a fmt violation in increment 1)
- Pre-existing (NOT introduced here): feature-gated `provenance_e2e`/`flow_red_team` test-compile errors also fail identically on origin/main.

No behavior change to any egress path.

### DIFF
--- a/crates/portcullis-core/src/declassify.rs
+++ b/crates/portcullis-core/src/declassify.rs
@@ -184,10 +184,7 @@ pub struct DeclassificationToken {
     /// or unbound token, which the value-binding apply check refuses. A hex
     /// string on the wire, like the signature — a 32-byte array is a cleaner
     /// commitment than a 32-element JSON number list for the governor endpoint.
-    #[cfg_attr(
-        feature = "serde",
-        serde(default, with = "commitment_hex")
-    )]
+    #[cfg_attr(feature = "serde", serde(default, with = "commitment_hex"))]
     pub content_commitment: [u8; 32],
     /// Ed25519 signature over the token's canonical form.
     /// Zero-filled if unsigned (for testing).
--- a/crates/portcullis-core/src/declassify.rs
+++ b/crates/portcullis-core/src/declassify.rs
@@ -171,6 +171,24 @@ pub struct DeclassificationToken {
     pub valid_until: u64,
     /// Human-readable justification (included in receipts).
     pub justification: String,
+    /// SHA-256 commitment to the exact content the token authorizes releasing —
+    /// the bytes the governor reviewed. Bound into the signed
+    /// [`Self::canonical_bytes`] (v3) so a token is valid ONLY for the value it
+    /// was minted over: at apply time the target node's recorded ingest content
+    /// hash must equal this, or the release is refused. This closes the
+    /// value-steering vector — where an adversary arranges for the node to hold a
+    /// different value than the one the governor signed for — which the
+    /// sink-scope binding (the *where*) does not address (the *which value*).
+    ///
+    /// `[0u8; 32]` (the [`Self::new`] default) means "no value binding": a legacy
+    /// or unbound token, which the value-binding apply check refuses. A hex
+    /// string on the wire, like the signature — a 32-byte array is a cleaner
+    /// commitment than a 32-element JSON number list for the governor endpoint.
+    #[cfg_attr(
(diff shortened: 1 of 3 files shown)


brandon-coproduct:
## Update: C4 fork RE-DECIDED after pressure-test — demote, not wording-fix

Brandon asked me to pressure-test the C4 call. I reversed it. The wording-fix (keep PROVED) was insufficient; C4 is now **demoted PROVED -> NOT-YET** (ratchet 5 -> 6, ledger green at 9/6).

Decisive reason — the PROVED row was overclaiming a live guarantee, **and vacuously**:
- The proven single-use token mechanism fires **nowhere in production** (tool-proxy `POST /v1/declassify` -> empty kernel `flow_graph` -> `NodeNotFound`; `nucleus-mcp` populates a graph but never applies a token). "The release is single-use" is vacuously true with zero releases — the anti-vacuity trap.
- The release path that **actually flips a shipping egress verdict** is a *different, unproven* mechanism: the k-of-n witness-threshold memory declassify (`nucleus-tool-proxy/src/memory.rs`, `memory_ifc_e2e.rs` asserts the flip). Not a single-use token; none of C4's proofs.

So the flagship row credited a proven-but-dormant mechanism while the live one is unproven — the exact inversion PROVED would hide. The single-use theorem stays fully credited in the note as a real kernel-API property; NOT-YET is the honest status of the *clause on the shipping path*. Re-earn = the graph-unification arc (Phase 2/3/4) under boot-a-real-pod.

Reversible in one commit if the owner disagrees.

PR #2268, opened 2026-08-12
C2 G4: live boot proves cross-pod isolation over real vsock — child C included, sibling B excluded (VERIFIED on KVM)

brandon-coproduct:
## What

The C2 capstone: a **booted** pod A, calling scoped `POD_LIST` over its **own real workload-API vsock**, is served a listing scoped to itself — sibling **B excluded** — while the operator sees both. That is cross-pod non-interference over the **real transport**, the one thing only a KVM host can show (the KVM-free gate proves the *filter*; unit tests use a `UnixListener` mock).

Three commits:
1. **Foundation** — `build-rootfs.sh` bakes `nucleus-podlist-probe`; `podlist-probe-pod.yaml` is pod A's spec (orchestrator + `enable_pod_mgmt`).
2. **Probe fix (found by the live boot)** — the probe runs as a *workload* spawned by the tool-proxy, so it does **not** inherit guest-init's `NUCLEUS_POD_ID` env. It now fetches its own id over the same socket-authenticated vsock (`FETCH_POD_CALLER_TOKEN` → `{caller_token, pod_id}`; `parse_pod_id`, 10 unit tests). Exactly the class of defect the boot harness exists to catch.
3. **Live gate** — `scripts/firecracker/podlist-boot-check.sh`.

## Verified on real Firecracker (Lima `nucleus-kvm`, aarch64)

```
A's booted POD_LIST ids = [A]
operator ids            = [A, B]
OK: A's vsock listing returned A and EXCLUDED sibling B — a strict subset of the
    operator view that sees both. Served over the real transport, not fail-open.
```

The gate sequences two pods (fixed `/run/firecracker.socket` ⇒ one boot at a time): B boots, its probe runs, B exits and **lingers** in the registry (so its absence from A's view is *exclusion*, not "never booted"), then A boots and asserts `A ∈ ids ∧ B ∉ ids ∧ {A,B} ⊆ operator ∧ ids ⊊ operator`.

The lineage-**inclusion** half (a child C ∈ A's view — the self-only tooth) stays proven KVM-free in `cross-pod-scoped-check.sh` (#2267) over the **same** `scope_to_caller` filter, so it isn't re-proven here.

## Scope

- Inputs taken as given (`KERNEL`/`ROOTFS`/`NODE_BIN`), like `boot-harness.sh`.
- **Remaining (mechanical):** wiring this as a **non-required x86_64 CI lane** in `quickstart-boot.yml` (build rootfs with `POD_SPEC=podlist-probe-pod.yaml`, run the check). Verified on aarch64 KVM; the x86_64 CI path is unverified in-session, so it lands as a follow-up rather than blind here.
- **C2 stays NOT-YET** until that lane is green and you promote the posture claim.

Not on automerge — touches the boot rootfs-build path; admin-merge after CI green.

https://claude.ai/code/session_0148ebbw4UXypRjhMw3xAQzj

### DIFF
--- a/crates/nucleus-podlist-probe/src/main.rs
+++ b/crates/nucleus-podlist-probe/src/main.rs
@@ -49,6 +49,14 @@ const VMADDR_CID_HOST: u32 = 2;
 const DEFAULT_WORKLOAD_API_PORT: u32 = 15012;
 /// A bound so a wedged host cannot make the probe hang past its drain window.
 const READ_TIMEOUT_MS: u64 = 2000;
+/// Poll POD_LIST a few times: a child or sibling created around the same instant
+/// as this pod may still be REGISTERING when the probe first fires. The node's
+/// filter excludes a non-lineage sibling on every poll, so a view that GROWS
+/// across polls can only mean a lineage member (a child) appeared — never a leak
+/// — so reporting the largest view is sound and races-free. Bounded well within
+/// the pod lifetime.
+const POLL_ATTEMPTS: u32 = 10;
+const POLL_INTERVAL_MS: u64 = 1000;
 
 const PASS_SENTINEL: &str = "NUCLEUS_PODLIST_PROBE: PASS";
 const FAIL_SENTINEL: &str = "NUCLEUS_PODLIST_PROBE: FAIL";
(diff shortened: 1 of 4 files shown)


brandon-coproduct:
**Update — the live proof now covers the full property, not just B-exclusion.**

Added a commit that upgrades the gate to **three pods** and proves lineage-**inclusion** live too, closing the same self-only vacuity hole over the real transport that #2267 closes KVM-free:

```
A(orch)=81a7cda2   C(child)=65d9ad5b   B(sib)=80f03a5b
A's POD_LIST ids = [A, C]        ← child C INCLUDED, sibling B excluded
operator ids     = [A, B, C]
OK — lineage-scoped (self + children), not self-only, not fail-open.  exit 0
```

- **Jailer** (the node default) gives per-pod sockets → A, C, B boot **concurrently**; C's parent is set to A via the `x-nucleus-parent-pod-id` header the node records for an operator create; C and B are created in parallel to register inside A's boot window.
- **Probe polls** `POD_LIST` and reports the largest scoped view — a sibling can never enter A's view, so a growing view is strictly more of A's own lineage, removing the boot-timing race without weakening the property.

Second KVM-only finding, fixed and documented: a long jailer chroot base overruns the **~108-char AF_UNIX limit** on the nested per-pod vsock socket → the pod 502s on health in a way that reads like an auth/routing bug. Keep the base short.

Verified on Lima `nucleus-kvm`, real aarch64 Firecracker.

PR #2284, opened 2026-08-13
ci(aeneas): bump Aeneas pin 2026.06.10 -> 2026.08.13-daa85d7

brandon-coproduct:
Bumps the two-month-stale Aeneas/Charon extraction pin across all 5 Aeneas workflows.

## What moves
- `AENEAS_RELEASE`: `nightly-2026.06.10` → `nightly-2026.08.13-daa85d7` (note the tag format now carries a `-<commit>` suffix).
- `CHARON_NIGHTLY`: **unchanged** at `nightly-2026-06-01`. The bundled Charon commit moves `9dd7f23` → `340b1af`, but I checked both commits' `rust-toolchain` and both build against `nightly-2026-06-01` — so the historical break vector (a toolchain mismatch) does not apply here.
- Prose pin in `crates/nucleus-github-oidc/lean/README.md` updated to the `08.13` commit for accuracy.

## Why this is safe to let CI judge
The Scoped-Aeneas jobs **re-extract against a fresh extraction** at the new pin and gate on:
- a clean axiom set (allowlist `[propext, Classical.choice, Quot.sound]`, no `*External` / `sorryAx`), with the non-vacuity guard that `#print axioms` actually ran;
- the `sorry`/`admit`/`native_decide` scan;
- the `lake build` under the pinned Lean toolchain.

So if the newer Aeneas emitted opaque axioms or failed to compile our extraction, these jobs would red — CI is the re-extraction oracle. The committed `generated/` dirs are a curated subset and the freshness diff in `aeneas.yml` is advisory (warns, rebuilds against fresh).

Not boot-affecting.

https://claude.ai/code/session_0148ebbw4UXypRjhMw3xAQzj

### DIFF
--- a/.github/workflows/aeneas-decide-pure.yml
+++ b/.github/workflows/aeneas-decide-pure.yml
@@ -49,7 +49,7 @@ permissions:
 env:
   # Pin to the Aeneas prebuilt nightly that does scoped extraction cleanly
   # (same pin the IFC scoped job uses). Update this AND regenerate together.
-  AENEAS_RELEASE: nightly-2026.06.10
+  AENEAS_RELEASE: nightly-2026.08.13-daa85d7
   # Charon needs a specific nightly for rustc-dev components.
   CHARON_NIGHTLY: nightly-2026-06-01
   # Scoped extraction roots — name them exactly as Charon sees the paths.
--- a/.github/workflows/aeneas-ifc-scoped.yml
+++ b/.github/workflows/aeneas-ifc-scoped.yml
@@ -60,7 +60,7 @@ permissions:
 env:
   # Pin to the Aeneas prebuilt nightly that generated the committed generated-ifc/.
   # Update this AND regenerate generated-ifc/ together.
-  AENEAS_RELEASE: nightly-2026.06.10
+  AENEAS_RELEASE: nightly-2026.08.13-daa85d7
   # Charon needs a specific nightly for rustc-dev components.
   CHARON_NIGHTLY: nightly-2026-06-01
   # The scoped extraction roots — name them exactly as Charon sees the paths.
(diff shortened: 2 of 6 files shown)


brandon-coproduct:
Closing — the bump to \`nightly-2026.08.13-daa85d7\` is a **net regression with no coverage upside**, so it should not land.

**Regression (whole-crate \`aeneas.yml\`):** under 08.13, Aeneas throws an uncaught \`Unimplemented\` / \`Aeneas.Errors.CFailure\` while translating the signature of \`core::slice::iter::Iter::find\` (a slice-iterator closure), transitively reached from \`crates/nucleus-ifc-kernel/src/flow.rs:1156\` and \`ifc_api.rs:825\`. The crash aborts the ENTIRE whole-crate translation, so Aeneas emits **zero .lean files**, and the fail-closed "a translator that produces zero output is a regression" guard reds. Under 06.10 the same extraction is a clean partial translation (exit 2, tolerated) and this job is green on main. The **scoped** \`--start-from\` pipelines (decide-pure, ifc-scoped, ck-policy, oidc-spiffe) DO pass on 08.13 — they don't pull in that iterator signature — but the legacy whole-crate liveness guard cannot.

**No coverage upside:** the constructs our hand-model notes cite as un-translatable (the 13-variant \`Operation\` enum, the bool-struct \`ExposureSet\`, \`ExposureLabel\`) are already machine-extracted at 06.10 (see \`generated-decide/\`). \`IFCLabel\` is a String-free struct-of-enums whose confidentiality clause is already extracted (\`extracted/ifc_confidentiality.rs\`). The real lever for more coverage is widening the extraction SCOPE + retiring the stale "Aeneas can't translate X" notes — both version-independent.

Staying on \`nightly-2026.06.10\`. Upstream regression worth revisiting on a future Aeneas release that fixes the \`Iter::find\` signature crash.

PR #2317, opened 2026-08-16
proof(gkat): finite-axiom completeness reduced to one open statement, with no uniqueness axiom

brandon-coproduct:
Attacks the "crystallization" route to GKAT finite-axiom completeness with the
uniqueness axiom eliminated. Two targets are refuted (kernel-checked), the
replacement architecture is built, and what remains is a single named statement.

## The chain

```
completeness_of_normalized :
  Normalizable → NormalCommonTarget → PullbackCovered → FiniteAxiomsCompleteBA
```

Everything between those is machine-checked. **No uniqueness axiom appears
anywhere.** `NormalCommonTarget` is discharged (`span_of_matched`), as is the
pullback and both its projections. `Normalizable` is Phase A pruning, not
started. `PullbackCovered` is the one genuine conjecture.

## Refutations (all kernel-checked)

- **`CommonSyntacticCollapse`** — the plan's own crystallization target — is
  **false**. GKAT automata are deterministic, so the quotient is *forced* rather
  than designed, and the forced one is unrealizable for an explicit pair.
  Completeness itself is untouched; only the route died.
- **`CommonTarget`** is **false**: `0` and `a ; 0` denote the same language, but
  one steps from its pseudostate and the other does not, and `star_bijection_init`
  says a cover preserves exactly that. Not confined to empty languages —
  `if b then (a;0) else c` vs `if b then 0 else c` dies the same way. Hence the
  productivity precondition.
- **`Nested ⟹ HasThompsonCover`** and **`ThompsonCofinal`** fail by search
  (6 counterexamples at three states, 84 at four). In both, *not one*
  counterexample is a pullback, while every crux pullback is covered.

## Proved

- `star_bijection` — an `InitCover` is a **covering map of graphs** (Stallings). Zero axioms.
- `unrollCover`, `cyclicCover` — W1-unrolling and loop-body doubling both refine,
  for arbitrary bodies. And `loop_doubling_provable`: doubling **is** derivable
  from the finite axioms, correcting a hand analysis that stalled exactly where UA lives.
- **The period law** (`GkatPeriodProofs`) — a covering map cannot shorten the
  cycle it sits over. `sorry`-free on `[propext]` alone.
- **The exit law** (`GkatLoopExitProofs`) — a GKAT `while` has exactly one exit
  condition, tested at the entry. That is the formal content of "GKAT has no `break`".
- The pullback, both projections, and the behavioural target with both legs of the span.

## Method

Every conjecture was gated by search before any Lean was written, which killed
four targets cheaply — including two I had recommended. The executable mirror of
the Lean Thompson construction lives in `lean/experiments/gkat-crystallization/`
(Rust + rayon, 56M automata at K=6).

The synthesis gate passes 273/273: `period(P) = lcm(period e, period f)` for every
crux pullback, so the degree the remaining construction needs is computable from
the two programs with no search.

## Verification

All 52 gated libraries build; no `sorryAx` anywhere in the proven tier.

https://claude.ai/code/session_01L1Bd7fNvWyBoTiDCXPM3JB


brandon-coproduct:
Closing: this work now lives in the private **`bcrisp/gkat-hierarchy`** repository.

The GKAT proof corpus was migrated out of nucleus earlier today — it is a pure-mathematics research program (the loop-nesting hierarchy, machine-checked for every `k`), and nucleus is a vendor-neutral secure runtime. Landing another **+179,418 lines across 100+ Lean files** here would pull the runtime repo further toward being a proofs monorepo, and would put every one of those files into the Lean CI tier on each run.

**Nothing is lost.** The branch `gkat-nary-and-hard-frontier` and all 780 commits remain on the remote, and the complete corpus — including the headline `depth_hierarchy_strict` result and the certificate families in this PR — is in `gkat-hierarchy` with its own dependency-free lakefile that builds in ~11s.

The GKAT theorems nucleus genuinely depends on stay where they are: the enforcement-core set audited by `portcullis-core-proven-lean.yml`, which as of #2331 actually runs and reports on all 90 of them.

Reopen if the migration needs revisiting.

PR #2336, opened 2026-09-01
feat(commerce): the receipt IS the payout claim — recompute-verifiable revenue share

brandon-coproduct:
## The receipt is the payout claim

Agent commerce standardised around **promises**. AdCP's own documentation puts attribution out of scope and defers it to third-party measurement; the research that builds verifiable agent receipts ([arXiv 2606.04193](https://arxiv.org/pdf/2606.04193)) names its own missing piece — *"adoption incentives: why services would implement receiver attestation."*

This answers that structurally: **a runtime emits verifiable evidence because that evidence is what its share is computed from.** There is no separate reporting channel to be honest on, and no operator ledger anyone has to trust.

## No new economics

A revenue split *is* a commons routing. `route_to_commons` is already the proven kernel — `Commons.lean`'s `routed_conserves`: a pool, basis-point shares that must sum to exactly 10 000, and a conservation theorem that the allocations neither skim nor over-allocate. Reusing it means the payout math **inherits a Lean proof** instead of acquiring a new trusted numeric path.

`verify_payout` is two stages, in this order:

1. the underlying clearing must itself re-derive (`verify_receipt`) — a payout over a fabricated price fails **here**, before the split is even examined;
2. the pool is **derived** from that clearing, never declared separately, and `route_to_commons` must reproduce the claimed allocations exactly.

Stage 1 is the load-bearing one. Letting a payout declare its own pool would leave the split honest and the number arbitrary.

Both stages are **mutation-tested**: each was removed in turn, and exactly one test reds — the right one.

## First producers for the economic lineage edges

`EdgeKind::{Bid, Allocation, Settlement, Dispute}` have existed as a schema since they landed, with **nothing constructing them**. `LineageEdge::allocation` and `::settlement` are the first producers, so a payout is recorded in the same lineage graph as the work that earned it.

The documented refund semantics are now exercised rather than only described: a refund is a `Settlement` with a negative signed amount whose parents include the edge it reverses, and the signed sum over a charge is zero exactly when the refund is full.

## Finding: the signature covers less of an economic edge than it looks like

`canonical_edge_bytes` signs `kind_tag(&kind)` — **the discriminant only**. A `Settlement` edge's `tx_ref` and `rail` are not covered, and neither is `attrs`. Two settlements differing only in rail-side transaction id produce byte-identical canonical bytes.

This is **pinned as a test rather than fixed**, because re-encoding would break every existing signature. The consequence for anything built on these edges: the authoritative binding is `content_hash_hex`, and the payout receipt behind it must carry the settlement reference itself. Verifying a payment against `attrs["amount_micro_usd_signed"]` or the edge's `tx_ref` would be a false guarantee.

## Surfaces

- `POST /v1/payout/verify` on `nucleus-verifier-service`
- `verifySignedPayout()` in `sdks/verifier-js` — a payee recomputes its own share **in the browser**, no server, no operator

A payout rides in `Projection::Economic` under inner kind `"payout"`, using the extension point that projection already documents (*"so other economic bodies can coexist under the same projection kind"*) — no change to `nucleus-receipt`'s sealed sum.

## What this does NOT claim

Recompute checks the split is the proven function of the **declared** inputs — the same honest boundary `verify_receipt` documents. `Attribution` (which workload surfaced which offer) is **bound into the signed content hash so it cannot be altered afterwards, but it is carried, not verified.** Proving it requires checking the workload's SPIFFE attestation independently, exactly as `MediationReceipt` already requires. The module docs say this in those words, and a test asserts that re-attributing a payout changes its hash *and still recomputes* — binding is not verification.

## Verification

- `wasm32-unknown-unknown` build verified: the wasm-pure default closure is unchanged (the envelope bridge stays behind the existing `envelope` feature).
- Tests: `nucleus-recompute` 63, `nucleus-lineage` 193, `nucleus-verifier-service` 96, `verifier-js` 19. Zero clippy across all four.
- Load-bearing cases: a *validly signed* but skimmed payout returns `mismatch:allocations`; a payout over a fabricated clearing returns `mismatch:seller_gross`; a share set that doesn't sum to 10 000 bps cannot be issued at all.

## Scope

Additive: one new module, two new edge constructors, one route, one SDK function. No existing behaviour changed. Settlement to a real rail is the next slice — this one stops at the claim.

### DIFF
--- a/crates/nucleus-lineage/src/edge.rs
+++ b/crates/nucleus-lineage/src/edge.rs
@@ -695,9 +695,8 @@ mod tests {
             "e".repeat(64),
         );
 
-        let amount = |e: &LineageEdge| -> i64 {
-            e.attrs["amount_micro_usd_signed"].parse().unwrap()
-        };
+        let amount =
+            |e: &LineageEdge| -> i64 { e.attrs["amount_micro_usd_signed"].parse().unwrap() };
         assert_eq!(amount(&charge) + amount(&refund), 0);
         assert!(
             refund.parents.contains(&charge.child),
--- a/crates/nucleus-lineage/src/edge.rs
+++ b/crates/nucleus-lineage/src/edge.rs
@@ -486,6 +486,83 @@ impl LineageEdge {
         }
     }
 
+    /// Construct an **allocation** edge: a market mechanism split `content_hash_hex`'s
+    /// payout among its payees.
+    ///
+    /// The economic `EdgeKind`s have existed since the schema landed with nothing
+    /// constructing them. This is the first producer, and it exists so a payout is
+    /// recorded in the same lineage graph as the work that earned it rather than in
+    /// a separate ledger somebody has to be trusted about.
+    ///
+    /// `content_hash_hex` MUST be `nucleus_recompute::payout::payout_content_hash_hex`
+    /// of the payout being recorded. It is taken as an argument rather than computed
+    /// here so `nucleus-lineage` keeps no dependency on the economic kernels — and,
+    /// more importantly, so the hash that is *signed* is the same one an independent
+    /// verifier recomputes. Binding via `content_hash_hex` and not `attrs` is
+    /// deliberate: `canonical_edge_bytes` signs the content hash and does NOT sign
+    /// `attrs`, so a binding placed in `attrs` would be a false guarantee.
+    pub fn allocation(
+        child: CallSpiffeId,
+        parents: Vec<CallSpiffeId>,
+        market_id: impl Into<String>,
+        mechanism: impl Into<String>,
+        content_hash_hex: impl Into<String>,
+    ) -> Self {
+        Self {
+            child,
+            parents,
+            kind: EdgeKind::Allocation {
+                market_id: market_id.into(),
+                mechanism: mechanism.into(),
+            },
+            content_hash_hex: Some(content_hash_hex.into()),
+            ts: Utc::now(),
+            attrs: BTreeMap::new(),
+            proof: None,
+            verifier_attestation: None,
+        }
+    }
+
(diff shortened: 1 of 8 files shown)


bcrisp:
### Follow-up worth its own change: the scoreboard's non-test filter misses compound `cfg(test)`

While fixing the `permissive_verify` ratchet on this PR I found a real bug in `scripts/exemplar-scoreboard.sh`.

The `nt()` filter strips `#[cfg(test)]` module bodies so test code doesn't inflate "production" counts. It matches only the **bare** form:

```awk
state==0 && $0 ~ /#\[cfg\(test\)\]/ { state=1; next }
```

So `#[cfg(all(test, feature = "x"))]` and `#[cfg(any(test, ...))]` are not stripped, and their bodies count as production. This is the same leak the filter's own comment already describes — *"the old filter dropped only the `#[cfg(test)]` attribute LINE… and repeatedly inflated permissive_verify"* — recurring in a new spelling.

Correcting the pattern to also match compound cfgs gives:

| metric | now | corrected |
|---|---|---|
| `permissive_verify` | 24 | **20** |
| `verify_calls_GUARD` | 53 | **49** |

It was miscounting **four** feature-gated test modules, three of them pre-existing and unrelated to this PR.

**I did not include the fix here, on purpose.** It drops `verify_calls_GUARD`, which the anti-gaming check correctly refuses without a deliberate re-baseline — and changing the measuring instrument inside the PR whose number it measures is the wrong way to do that. It deserves a change where the re-baseline *is* the point and can be reviewed as such.

The baseline move in this PR (`permissive_verify` 22→24, `verify_calls_GUARD` 51→53) is therefore honest under the **current** ruler, and one of those two counts is a known artifact that the follow-up removes.

PR #2357, opened 2026-09-01
release(spec): pin the guest to v2.2.0 and raise the floor above 2.1.0

brandon-coproduct:
## The pinned guest has not been bootable for five weeks

`GUEST_RELEASE` has been `2.1.0` since **2026-07-30**. #2214 landed **2026-08-08** and changed how the guest is approved: a current `nucleus-node` sends `nucleus.approval_pubkeys` and no longer sends `nucleus.approval_secret` — which the 2.1.0 rootfs's guest-init still requires.

```text
nucleus-guest-init error: missing approval secret (set nucleus.approval_secret …)
Kernel panic - not syncing: Attempted to kill init! exitcode=0x00000000
```

The guest exits; because it is PID 1 the kernel panics. `setup` installed this anyway.

## This is the failure the floor already exists for

`GUEST_RELEASE_FLOOR` was introduced by #2110: everything up to 2.0.2 ships a rootfs with no CA bundle, the tool-proxy panicked **as PID 1**, and the floor made `setup` refuse rather than install a pod that cannot boot.

Different cause, identical outcome, and the remedy is the one already in the file — **2.1.0 goes below the floor.**

| | before | after |
| --- | --- | --- |
| `GUEST_RELEASE_FLOOR` | 2.1.0 | **2.2.0** |
| `GUEST_RELEASE` | 2.1.0 | **2.2.0** |

## Ordering, and its cost

Bumped **before** the tag is cut, matching how 2.1.0 was bumped from its RC in the change released as 2.1.0. That ordering has a real cost and it is now documented on the constant rather than left implicit: **until the `v2.2.0` assets are built, `setup` points at a release that does not exist.**

The alternative is worse. Tag first and bump after ships a release whose CLI pins the *previous* release's guest — precisely the skew being closed.

**So this PR must be followed promptly by the `v2.2.0` tag.** `release.yml` fires on `v*` and builds the guest binaries and rootfs; `crates-publish.yml` fires on `crates-v*` and is untouched by this, so nothing reaches crates.io.

## Verification

- 58 `nucleus-spec` tests pass.
- Two tests used 2.1.0 as their example of an **acceptable** version and went red. They test suffix parsing, not that version, so they now use 2.2.0.
- The useful half: **2.1.0 was added to `a_prerelease_of_a_refused_version_is_still_refused`.** The floor rise now has teeth — lowering it back fails an assertion instead of passing quietly. Without that, raising the floor would be a constant nothing checks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,10 +73,11 @@ nucleus setup --install-deps   # installs Lima if missing, provisions the VM,
                                # POD and asserts what the guest did
 ```
 
-Guest artifacts come from the pinned release **v2.1.0**, the first one able to
-boot a pod at all — everything up to 2.0.2 ships a rootfs with no CA bundle, on
-which the guest panics as PID 1, and `tier2_artifacts::GUEST_RELEASE_FLOOR`
-refuses those rather than installing one. **Measured 48.7 s** from a deleted VM to
+Guest artifacts come from the pinned release **v2.2.0**, the first whose rootfs
+matches a current node. `tier2_artifacts::GUEST_RELEASE_FLOOR` refuses anything
+older rather than installing a pod that cannot boot — everything up to 2.0.2
+ships a rootfs with no CA bundle, on which the guest panics as PID 1, and 2.1.0
+predates the change to how the guest is approved, which panics the same way. **Measured 48.7 s** from a deleted VM to
 a booted pod, with Sigstore build provenance verified on every downloaded
 artifact.
 
--- a/crates/nucleus-spec/src/tier2_artifacts.rs
+++ b/crates/nucleus-spec/src/tier2_artifacts.rs
@@ -85,24 +85,38 @@ pub const RELEASE_REPO: &str = "coproduct-opensource/nucleus";
 /// Both halves of that fix — the fallible constructor and `build-rootfs.sh`
 /// installing a bundle — landed after 2.0.2.
 ///
+/// The same thing has now happened a second time, one release up. #2214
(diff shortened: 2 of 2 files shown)


brandon-coproduct:
**Verified the bump propagates and that the window is non-fatal.**

`pins.json` is generated from this constant (`cargo run -p nucleus-cli -- verify --pins`), so there is a single source of truth and no second copy to drift. On this branch it now resolves:

```
2.2.0  nucleus-rootfs-2.2.0-aarch64.ext4.gz
2.2.0  nucleus-node-2.2.0-aarch64-unknown-linux-musl.tar.gz
2.2.0  nucleus-cli-2.2.0-aarch64-unknown-linux-musl.tar.gz
2.2.0  nucleus-rootfs-2.2.0-x86_64.ext4.gz
2.2.0  nucleus-node-2.2.0-x86_64-unknown-linux-musl.tar.gz
2.2.0  nucleus-cli-2.2.0-x86_64-unknown-linux-musl.tar.gz
```

Those are exactly the six names `release.yml` publishes — checked against the **real** asset list of v2.1.0, which carries the same pattern (`nucleus-rootfs-2.1.0-aarch64.ext4.gz`, etc.).

**The bump-before-tag window does not redden CI.** `quickstart-boot.yml` already distinguishes the two states deliberately — "pinned release not published yet" is a `::notice::` with `exit 0`, while "release exists but an asset is gone" is an `::error::`. That is precisely the distinction this ordering needs, and it was already there.

So the only thing standing between this landing and the pin being real is the `v2.2.0` tag.

Issue #2396, opened 2026-09-03
nucleus setup silently replaces locally-built binaries, and the symptom is a bare 401


brandon-coproduct:
On a host where you have just installed your own build, `nucleus setup` overwrites `/usr/local/bin/nucleus` and `/usr/local/bin/nucleus-node` with the released artifacts. Nothing says it did. The next signed request fails with `401 invalid signature`, which points at your credentials rather than at the swap.

## What happened

```
21:59  built  nucleus / nucleus-node   (from main)
22:03  sudo cp ... /usr/local/bin/     (installed my build)
22:03  nucleus setup --skip-verify     (silently replaced both with v2.2.0)
```

```
nucleus      installed=5c22ac7abb26c1ae  built=4848d1cf18bd23d4  DIFFERENT
nucleus-node installed=d1c1ed59ae969544  built=5a62d64897140bdc  DIFFERENT
```

## Why it presents as an auth bug

v2.2.0 predates nonce-based replay protection:

```
strings /usr/local/bin/nucleus-node | grep -c x-nucleus-nonce   -> 0   (installed, v2.2.0)
strings target/.../nucleus-node     | grep -c x-nucleus-nonce   -> 1   (built, main)
```

A client built from main signs `ts.actor.nonce.body` and sends `x-nucleus-nonce`. The old node ignores the header and verifies `ts.actor.body`. The HMACs differ, so every request returns:

```json
{"error":"auth error: invalid signature"}
```

The secret is correct and identical on both sides. Confirmed by capturing both clients on the wire: the v2.2.0 CLI sends no `x-nucleus-nonce` header, a main-built client does.

This cost a long debugging detour — the secret was checked against the file, against `/proc/<pid>/environ`, and re-derived several ways before the wire capture showed the two clients disagreeing about the protocol rather than the key.

## Two things worth fixing

1. **Say what it is doing.** `nucleus setup` should report that it is replacing an existing binary, and ideally skip or prompt when the installed one is newer than the release it is about to write.
2. **Make version skew diagnosable.** A node that receives `x-nucleus-nonce` but does not understand nonces cannot verify that signature, and could say so — "signed with a nonce, which this node does not support" is a one-line diagnosis instead of an afternoon. A version handshake, or including the node version in the 401 body, would do it.

The second half matters beyond this case: any protocol change to the signing scheme will present as `invalid signature` to whoever is running a mismatched pair.

Issue #2395, opened 2026-09-03
A guest_cid other than 3 panics the guest, and nothing says so


brandon-coproduct:
Setting `spec.vsock.guest_cid` to any value other than `3` kills the pod. The spec field is accepted, written into `firecracker.json`, and then the guest cannot bind to it.

## Measured

Same spec, same host, only `guest_cid` varied (aarch64, KVM, node + rootfs both built from main):

| guest_cid | result |
| --- | --- |
| 3 | boots, full startup trace |
| 4 | `EADDRNOTAVAIL`, PID 1 exits, kernel panic |
| 100 | `EADDRNOTAVAIL`, PID 1 exits, kernel panic |

Guest console for `guest_cid: 4`:

```
nucleus-startup-phase vsock_bind=1ms at=396ms
Error: Io(Os { code: 99, kind: AddrNotAvailable, message: "Address not available" })
[    2.128189] Kernel panic - not syncing: Attempted to kill init! exitcode=0x00000100
```

`firecracker.json` for that pod really does carry `"guest_cid":4`, so the value reaches Firecracker; the failure is the guest binding a listener to it.

## Where

`crates/nucleus-tool-proxy/src/pod_mgmt.rs:395` takes the CID from the spec and `:446` binds a **listener** to it:

```rust
let cid = args.vsock_cid
    .or_else(|| spec.spec.vsock.as_ref().map(|v| v.guest_cid))
    .unwrap_or(tokio_vsock::VMADDR_CID_ANY);
let listener = tokio_vsock::VsockListener::bind(VsockAddr::new(vsock.cid, vsock.port))?;
```

A guest can only bind its own local CID (or `VMADDR_CID_ANY`, which is what the `unwrap_or` already falls back to). Taking the CID from the spec is what breaks it.

## Worth noting: concurrency is fine

vsock CIDs are per-VM, not host-global, so this is not a limit on parallelism — **three pods all using `guest_cid: 3` booted concurrently without complaint**. The bug only bites callers who assume they must hand each pod a distinct CID, which is a natural assumption to make from the spec field existing.

## The diagnosis is the other half

The node reports this as:

```
400 driver error: proxy health check timed out after 30s: the guest ANSWERED with
`HTTP/1.1 502 Bad Gateway` ... Kernel panic - not syncing: Attempted to kill init!
```

[...]

PR #2397, opened 2026-09-03
feat(perf): measure real pod concurrency instead of estimating it

brandon-coproduct:
Adds `nucleus-perf podburst`: launch N real pods at once against a running node, over the same signed HTTP API `nucleus node create` uses, and report what the host actually did.

## Measured on aarch64 / KVM / 4 vCPU / 7.9 GiB

| N | started | failed | submit p50 | time-to-running p50 | RSS each |
| --- | --- | --- | --- | --- | --- |
| 1 | 1 | 0 | 5.6s | 5.6s | 50 MiB |
| 5 | 5 | 0 | 10.7s | 10.9s | 50 MiB |
| 10 | 0 | **10** | 31.9s | — | 50 MiB |
| 25 | 0 | **25** | 33.8s | — | — |
| 50 | 5 | **45** | 67.7s | 112.2s | 50 MiB |

Largest clean burst: **5 pods**.

**Memory is not the wall.** Each pod is resident in ~50 MiB against 512 MiB configured — a 10x over-estimate. 100 pods would need ~5 GiB resident on a host with 7.9 GiB. Anyone sizing capacity by multiplying the configured size buys roughly ten times the hardware they need.

**The wall is a fixed 30s timeout under CPU contention.** Every failure at N>=10 is `proxy health check timed out after 30s`, not an allocation failure.

## Why the columns are split the way they are

Four choices, each guarding a way this measurement is normally wrong:

* **Submit latency and time-to-running are separate.** The first is the node accepting the request, the second includes the microVM booting. Collapsed, you cannot tell which half is slow — and here they diverge sharply at N=50 (67s vs 112s).
* **Configured and resident memory are separate.** See above; this is the whole capacity-planning story.
* **Pods that never reach Running are counted, not dropped.** A burst that "succeeds" quickly because most of it failed is the failure this harness exists to catch, so `failed` is a first-class column and the summary names the first level where it goes non-zero.
* **Percentiles, not a mean.** Under contention the tail is the story.

## Two bugs found by building it

* #2395 — a `guest_cid` other than 3 panics the guest. The harness varied CIDs per pod on the assumption they had to be unique; every burst died until that was removed. vsock CIDs are per-VM, so pods can and should share one.
* #2396 — `nucleus setup` silently replaced locally-built binaries with v2.2.0, and the resulting protocol skew presented as `401 invalid signature`.

The harness also caught a bug in itself: with `http_status_as_error(false)`, a 401 parsed cleanly as JSON, found no `pods` key and returned an empty list — so a failed auth read as "the node is up and idle". Status is now checked on every call, which is what turned an afternoon of guessing into a one-line diagnosis.

No new dependencies: `ureq` and the existing `nucleus-client` signing helper. Dev-only, `publish = false`, like xtask. Edition 2021 so it builds on main.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7264,6 +7264,20 @@ dependencies = [
 ]
 
 [[package]]
+name = "nucleus-perf"
+version = "1.0.0"
+dependencies = [
+ "anyhow",
+ "clap",
+ "hex",
+ "nucleus-client",
+ "portcullis",
+ "serde_json",
+ "serde_yaml",
+ "ureq",
+]
+
+[[package]]
 name = "nucleus-permission-market"
 version = "1.0.0"
 dependencies = [
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,5 +1,6 @@
 [workspace]
 members = [
+    "crates/nucleus-perf",  # Dev-only: pod concurrency + start-latency measurement
     "crates/portcullis",  # Policy: quotient lattice types
     "crates/nucleus-policy-kernel", # PCA: reference monitor (decide / governance_monotone)
     "crates/nucleus-policy-cert",   # PCA: recompute-verifiable certificates (R1)
--- a/crates/nucleus-perf/Cargo.toml
+++ b/crates/nucleus-perf/Cargo.toml
@@ -1,0 +1,21 @@
+[package]
+name = "nucleus-perf"
+version.workspace = true
+edition.workspace = true
+rust-version.workspace = true
+license.workspace = true
+authors.workspace = true
+# Dev-only measurement harness — never published.
+publish = false
+readme = "README.md"
+
+[lints]
+workspace = true
+
+[dependencies]
+anyhow = { workspace = true }
+clap = { workspace = true }
+serde_json = { workspace = true }
+serde_yaml = { workspace = true }
+ureq = { workspace = true }
+nucleus-client = { path = "../nucleus-client" }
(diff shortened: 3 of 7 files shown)


brandon-coproduct:
Iteration 2 ran on `nucleus-kvm`. It **failed**, and the reason is worth keeping.

```
pod c0fdfe1d up in 3042 ms, proxy http://127.0.0.1:42723

check                       status      ms  verdict
glob finds sandbox files       200     197  ok
read audit                     403     134  FAILED
      {"error":"access denied: path 'audit' blocked by policy","kind":"path_denied"}
forbidden read refused         403      50  ok
uncredentialed refused         403      52  ok
```

**There is no file in the pod sandbox a `codegen` pod may read.** The root contains only `audit`, and reading it is refused as sensitive — correctly. So the read path cannot be demonstrated until something readable exists. Writing one first is also blocked: `write_files` is `LowRisk` in `codegen`, so an unapproved write returns `approval_required` and needs `/v1/approve`.

Two rubric premises were wrong and are corrected in this push:

- iteration 2 asked for a host-side checksum; the sandbox is a disk image, not a host mount, so the host can neither place a file in it nor read one out
- iteration 3 had the same problem and omitted the approval prerequisite

Ledger row 2b records the unblock: bake a readable fixture into the rootfs (preferred — it keeps read and write independent) or implement the approval path.

What did pass: **3,011 ms cold boot on an idle host** against the 5,580 ms measured under load, glob served in 197 ms, and both refusals behaved — including the deliberately uncredentialed `web_fetch`, which is what makes the admission gate non-decorative.

Issue #2400, opened 2026-09-03
A path refusal says "blocked by the path lattice" without saying which rule


brandon-coproduct:
A pod refused two reads, and neither message is actionable:

```
access denied: path 'audit' blocked by policy          kind=path_denied
access denied: path 'perf-1756869037.txt' blocked by policy  kind=path_denied
```

Both paths are permitted by the standalone sensitive-path lattice. Probed directly:

```
PROBE          audit -> can_access=true
PROBE        ./audit -> can_access=true
PROBE     perf-1.txt -> can_access=true
PROBE    .ssh/id_rsa -> can_access=false     <- the only one actually blocked
PROBE      README.md -> can_access=true
```

So the refusal is coming from the pod's *effective* policy — the kernel decision, including work_dir sandboxing — and not from `PathLattice::block_sensitive()`. That is legitimate. The problem is that the message points the reader at the wrong place: an operator who reads "blocked by the path lattice" goes and inspects the blocklist, finds the path allowed there, and has learned nothing.

## Where it comes from

`DenyReason::PathBlocked` (`portcullis/src/kernel.rs:209`) carries only the path:

```rust
PathBlocked {
    /// The blocked path.
    path: ...,
}
```

and the mapping hardcodes a reason string (`nucleus-tool-proxy/src/mediation.rs:46`):

```rust
DenyReason::PathBlocked { path } => ApiError::Nucleus(NucleusError::PathDenied {
    path: std::path::PathBuf::from(path),
    reason: "blocked by the path lattice".to_string(),
}),
```

The kernel knows which constraint fired — whether the path escaped work_dir, matched a blocked glob, or failed an allowlist — and discards that before the caller sees it.

## Suggested fix

Give `PathBlocked` a rule discriminant (escaped sandbox / matched blocked pattern `<glob>` / not in allowlist) and render it in the message. Two callers already destructure this variant exhaustively, so the change is contained, and `mediation.rs` explicitly prefers exhaustive matching over a catch-all for exactly this reason.

## Related

Same family as #2398, where the same refusals were additionally being run through a redactor that replaced every `/` with `[home]`. With that fixed the messages are legible; they are still not informative.

Found while checking the read path for the microVM-per-tool-call rubric — the whole of iteration 2 is currently blocked on not being able to tell why a read is refused.

brandon-coproduct:
[...]

PR #2401, opened 2026-09-03
feat(portcullis): let the path lattice say WHICH rule refused a path

brandon-coproduct:
A pod refused a read with:

```
access denied: path 'audit' blocked by policy    kind=path_denied
```

An operator reads that, opens the blocklist, and finds `audit` permitted. Probed directly:

```
audit -> can_access=true      ./audit -> true      perf-1.txt -> true
.ssh/id_rsa -> can_access=false                    <- the only real block
```

The refusal was genuine and came from the pod's effective policy, so the message is not *wrong* — it is unactionable. It names a **layer**, not a **rule**, and the three things that layer can mean have three different fixes: the path left the sandbox, a blocked glob matched, or an allowlist did not include it.

## What this adds

`PathLattice::deny_reason` returns a `PathDenial` carrying the specific evidence — the sandbox root that was escaped, **the exact glob that matched** (there are ~30 in the default set, so "a pattern matched" is not a diagnosis), or how many allow patterns were considered.

`can_access` is now defined as `deny_reason(path).is_none()`, so there is one decision procedure and the boolean is a view of it. A second function that merely *explains* would be a second thing to keep in sync, and it would drift.

## Scope

Purely additive — no existing signature changes, no match site needs updating. Threading it through `DenyReason::PathBlocked` so it reaches the HTTP caller is the follow-up (#2400). This is the part the kernel needs first, and it stands on its own for anyone holding a lattice.

## Tests

Five, two of which exist to stop the obvious cheats:

- **an allowed path must report no denial** — otherwise a `deny_reason` that always returned `Some` satisfies the other three and denies everything through `can_access`
- **`can_access` and `deny_reason` are checked to agree** across four lattices × six paths, pinning the relationship rather than trusting it

1110 lib tests pass, whole crate green, rustfmt clean. The single clippy warning in `tests/adversarial.rs` is pre-existing on main and untouched.


brandon-coproduct:
Pushed a fix to this branch before it could ship.

`PathDenial::EscapesSandbox` was rendering the `work_dir` into its message. That string goes to whoever asked, and inside a pod that is the untrusted guest — so the sandbox root, which is host layout, would have travelled to it the moment #2400 threads these through to the HTTP caller.

The obvious defence does not apply. The tool-proxy's only production sanitizer call is:

```rust
validation::sanitize_error_message(&self.to_string(), None)
```

so the `[sandbox]` redaction branch never runs (filed as #2402). The generic absolute-path fallback only rewrites runs longer than five characters, and `/work` is exactly five.

The field stays on the value — the host still wants it in a log. Only the rendering withholds it. Structured detail for the side that already knows; a bare reason for the side that should not.

Two tests, deliberately pulling against each other:
- the rendered escape must **not** contain the sandbox root
- the rendered block **must** name the glob that matched — that is the whole point, and withholding everything would be as unactionable as the "blocked by policy" message this replaces

1112 lib tests green, clippy clean.

brandon-coproduct:
Extended this PR to close the whole chain #2400 described — the lattice knew why it refused a path, and every layer above it discarded that, one at a time.

**The last link was the smallest and the worst:**

```rust
#[error("access denied: path '{path}' blocked by policy")]
PathDenied { path: PathBuf, reason: String }
```

`reason` is captured at **fifteen** construction sites and rendered at **none**. `sandbox.rs` passes the underlying IO error; the mediation layer passed "blocked by the path lattice" — and the formatter replaces all of it with "blocked by policy". That is verbatim the message a live pod returned, and it is why a refused read could not be diagnosed at all.

**The rest:**
- `DenyReason::PathBlocked` carries an optional `PathDenial`, `serde(default)` because `DenyReason` is a tagged wire format — a payload written before this field existed must still deserialize rather than erroring at a security boundary
- the kernel asks `deny_reason` instead of `can_access` — same decision procedure, one is a view of the other, so no verdict changes and no second check appears
- the tool-proxy and MCP paths render it

**What reaches the guest is deliberately asymmetric,** informed by [Causality Laundering: Denial-Feedback Leakage in Tool-Calling LLM Agents](https://arxiv.org/pdf/2604.04035): a matched glob is named, because an agent can map the blocklist by probing paths whatever we tell it — naming it saves round trips without granting anything. The sandbox root is withheld, because probing cannot reveal host layout. This matters more than it looks: the tool-proxy's only production sanitizer call passes `None` for the sandbox root (#2402), so nothing downstream would redact it.

Tests: the glob must appear, the root must **not**, and `denial: None` must still produce the old sentence — that last one so "render nothing" cannot pass.

nucleus 54, portcullis 1112, mcp 53, tool-proxy 361, all green. The two unused-import warnings in portcullis test files are pre-existing on main (verified against main with the same flags) and untouched.

Issue #2402, opened 2026-09-03
The [sandbox] redaction branch never runs in production


brandon-coproduct:
`sanitize_error_message` takes a `sandbox_root` so it can replace it with `[sandbox]`. The tool-proxy's only production call passes `None` (`nucleus-tool-proxy/src/main.rs:1305`):

```rust
// Sanitize error message to prevent information disclosure
let sanitized_error = validation::sanitize_error_message(&self.to_string(), None);
```

So that branch is exercised only by the module's own tests and by its doc example, both of which pass `Some(...)`:

```rust
/// let sanitized = sanitize_error_message(msg, Path::new("/var/sandbox/abc123"));
/// assert_eq!(sanitized, "failed to read [sandbox]/secrets/token.txt");
```

The documentation therefore describes behaviour that never happens on the live path. A reader — or a future author adding an error that mentions the sandbox root — would reasonably assume they are covered.

## What actually catches it today

Only the generic `sanitize_absolute_paths` heuristic, which redacts a run of path characters **longer than five**:

```rust
if path_len > 5 {
```

A sandbox root of `/work` is exactly five and passes through verbatim. Longer roots are caught, so this is not currently leaking in the common case — it is a gap that depends on the length of a path nobody chose with this in mind.

## Suggested fix

Either pass the real sandbox root at the call site so the specific branch does the work it was written for, or delete the parameter and the doc example so the contract matches the behaviour. The present state is the worst of the three: an unused safety argument that reads as though it is protecting something.

Related: #2398 fixed two other faults in this same function (a `HOME=/` that rewrote every separator, and a relative path redacted as an absolute one). Found while designing the denial messages for #2400 — the rendering there deliberately withholds the sandbox root rather than relying on this.

PR #2403, opened 2026-09-03
docs(perf): record what iteration 2b actually found

brandon-coproduct:
Iteration 2b probed three path forms against a live pod. None was accepted, and the answers were more informative than a pass would have been:

```
audit         -> 403 path_denied
./audit       -> 403 path_denied
/work/audit   -> 403 sandbox_escape
```

The third line means **`/work` is not the sandbox root**, even though it is the spec's `work_dir`. The rubric assumed otherwise.

Probing the lattice directly shows it *permits* all three — `audit`, `./audit`, `perf-1.txt` all `can_access=true`, with only `.ssh/id_rsa` blocked. So the refusal comes from the pod's effective policy rather than the sensitive-path blocklist, and the message names a **layer** instead of a **rule**. That is #2400, and it is what actually blocks iteration 2: the read cannot be diagnosed, so it cannot be fixed.

## Row 2c revises the unblock

The earlier plan preferred baking a readable fixture into the rootfs. The measured position is different: the write path needs `/v1/approve`, which is Ed25519 over `{round}.{timestamp}.{actor}.{body}` with a drand round (`nucleus-tool-proxy/src/auth.rs:396`), and the approver private key lives host-side at `/var/lib/nucleus/state/approval_signing_key.der`. So the harness can legitimately act as the approver, rather than the production rootfs acquiring a test fixture it would carry forever.

## Blockers table

Adds #2400 and #2402. The latter matters for the design: the `[sandbox]` redaction branch never runs in production, so a denial message must withhold the sandbox root itself rather than trust a sanitizer that is not looking.

The FAIL row is kept rather than overwritten — what did not work is the useful part of this file.


brandon-coproduct:
Follow-up measurement, now that #2401 renders the real reason. This sharpens row 2b considerably — I cannot push it here without dequeuing this PR, so recording it for the next ledger update.

Rebuilt the guest and host from the #2401 branch and re-ran the pod scenario:

```
read audit   403   {"error":"access denied: path 'audit': Is a directory (os error 21)","kind":"path_denied"}
```

**`audit` was never a policy denial. It is a directory.** The `reason` field carried `Is a directory (os error 21)` the whole time and the formatter replaced it with "blocked by policy" — so an operator chasing a security refusal was chasing an `EISDIR`.

Widening the glob from `*` to `**/*` returns **exactly one match**, that same directory. So the pod sandbox contains **no readable file at all**, which is why every path form failed. Row 2b guessed at this; it is now measured, with the cause named.

Consequence for the rubric: iteration 2 genuinely cannot pass by reading something that already exists — there is nothing to read. It has to create a file first, which is row 2c (the `/v1/approve` path, Ed25519 over `{round}.{timestamp}.{actor}.{body}` with a drand round, approver key host-side). The earlier "prefer a baked-in fixture" plan is now clearly the weaker option: the sandbox being empty is correct behaviour for a bare pod, not a defect to paper over.

Also visible in that output: `path '.[home]audit'` still appears, because this build predates #2398 merging. Not a regression.

PR #2405, opened 2026-09-03
feat(client): an Ed25519 approval signer, because the HMAC one cannot approve

brandon-coproduct:
`nucleus-client` offers `sign_approval_headers`, which signs with **HMAC**. A Firecracker pod never accepts it.

The node puts `nucleus.approval_pubkeys` on every pod's kernel cmdline, and the proxy picks its tier like this:

```rust
} else if is_approval_path && has_approval_pubkeys {
    AuthTier::ApprovalEd25519Drand
} else if is_approval_path {
    AuthTier::ApprovalHmacDrand
}
```

The HMAC tier is reachable **only when no approver keys are configured**, which is never true for a Firecracker pod. So the crate shipped a function that looks like the way to approve an operation, cannot approve one on a default deployment, and had no Ed25519 counterpart to reach for instead.

Found while trying to write a file from the rubric harness: `write_files` is `LowRisk` under `codegen`, so an unapproved write is refused with `approval_required`, and there was no way to produce an approval the pod would take.

## What changed

The message is unchanged — `{round}.{timestamp}.{actor}.` then the body, exactly what `build_drand_message` already produced — so **only the primitive differs**. The drand round is clock-derived (`expected_round_for_timestamp`), not fetched, which is what makes this usable from a host with no route to `api.drand.sh`.

## The tests live in nucleus-tool-proxy on purpose

Testing a signer against a message the test rebuilds itself proves only that the signer agrees with the test. These run against the real `verify_http_with_ed25519_drand`, so a disagreement about message bytes, header names, or signature encoding fails **here** rather than on a live pod.

- the client's signature verifies
- a key the operator did not configure is refused — without which a verifier that accepted anything would pass the first test
- **the HMAC helper does not satisfy the Ed25519 verifier**, pinning the exact gap this closes so it cannot quietly reopen

`Cargo.lock` is committed: `ed25519-dalek` is new to nucleus-client and CI builds with `--locked`.

nucleus-client 41, nucleus-tool-proxy 368, clippy and rustfmt clean. This unblocks rubric row 2c (write a file, then read it back) — the harness wiring follows separately.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7270,6 +7270,7 @@ version = "1.0.0"
 dependencies = [
  "anyhow",
  "clap",
+ "ed25519-dalek 3.0.0",
  "hex",
  "nucleus-client",
  "portcullis",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6693,6 +6693,7 @@ name = "nucleus-client"
 version = "1.0.0"
 dependencies = [
  "drand-verify",
+ "ed25519-dalek 3.0.0",
  "hex",
  "hmac 0.13.0",
  "reqwest 0.13.4",
--- a/crates/nucleus-client/Cargo.toml
+++ b/crates/nucleus-client/Cargo.toml
@@ -15,6 +15,9 @@ categories = ["authentication", "development-tools"]
 hmac = { workspace = true }
 sha2 = { workspace = true }
 hex = { workspace = true }
+# Approvals on a Firecracker pod are verified with Ed25519, not HMAC -- see
+# `sign_approval_headers_ed25519`.
+ed25519-dalek = { workspace = true }
 # v4 UUIDs are the per-request nonce that lets an otherwise-identical request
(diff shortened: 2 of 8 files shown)


brandon-coproduct:
Validated end to end against a live pod, and the signer works:

```
unapproved write refused       403   ok     <- the gate itself
approval accepted              200   ok     <- this PR, verified on a real pod
```

`/v1/approve` returning **200** means the drand round, the message construction, and `verify_http_with_ed25519_drand` all agree against a real Firecracker pod — not just against the interop test in this PR. The HMAC helper could not have produced that.

Also pushed the harness half (`--approval-key`): write → approve → retry → read back, with the unapproved write treated as a **check** rather than a step to route around.

The retry is still refused (#2406), but that failure is downstream of authentication and does not implicate this change — four hypotheses tested and refuted so far, so I filed it with the evidence rather than guessing a fifth time.

PR #2408, opened 2026-09-03
fix(node): a pod that fails to boot leaked its network namespace

brandon-coproduct:
Measured on a host with **zero** live Firecracker processes:

```
$ ip netns list
nuc-f45b3cc8
nuc-59c6d719
```

Those are exactly the two pods whose guests **kernel-panicked** during a `guest_cid` experiment (#2395). Every pod that booted cleanly reaped its own namespace; the two that died did not.

189 pod state directories on the same host and only these two namespaces — so this is not "cleanup never runs", it is **"cleanup runs on the paths somebody remembered"**.

## Where it leaks

`spawn_firecracker_pod` creates the namespace early and can fail at a dozen later points: config serialisation, Firecracker spawn, the proxy health wait, a guest that panics before it answers. The early arms each call `cleanup_netns` explicitly. The later ones do not — `return Err(ApiError::Driver("config serialize failed: ..."))` is one, and the health-wait failure that produced the two namespaces above is another.

## The fix does not enumerate error paths

Enumerating them is the thing that failed. `NetnsGuard` reaps on **any** unwind out of the creation path and stands down explicitly at the single success return, so a `?` added later by someone who never reads the file still reaps.

Two details that matter:

- the guard is declared **outside** the `if netns_plan.create_netns` block. Scoped to that block it would drop while the pod was still booting and delete a **live** namespace — worse than the leak it prevents
- `disarm` takes `self`, so a disarmed guard cannot be re-armed or dropped-armed by accident: the type rules out the dangerous case rather than a comment asking callers to avoid it

`Drop` cannot await, which costs nothing: `cleanup_netns` is a blocking `Command` and its `async` is decorative, so the guard runs the same command. Existing explicit cleanups now run before the guard's, and a second `ip netns del` is a no-op on an absent namespace.

## Scope

Rubric **iteration 8** (no leaked microVM, netns, cgroup or state dir). Also measured on the same host: **0** leaked cgroups, **0** leaked taps. The 189 state directories are audit material (`lifecycle.log`, receipts) and are a separate retention question, not this leak.

390 tests pass; clippy and rustfmt clean.

### DIFF
--- a/crates/nucleus-node/src/main.rs
+++ b/crates/nucleus-node/src/main.rs
@@ -2206,9 +2206,18 @@ async fn spawn_firecracker_pod(
                 "network policy requires --firecracker-netns=true".to_string(),
             ));
         }
+        // Declared OUT here on purpose. The namespace is made inside the block
+        // below but must survive until this function succeeds, so a guard scoped
+        // to that block would reap a live pod's namespace the moment the block
+        // ended — worse than the leak it exists to prevent.
+        let mut netns_guard: Option<net::NetnsGuard> = None;
         if netns_plan.create_netns {
             let name = net::netns_name(id);
             net::create_netns(&name).await?;
+            // Armed from here to the single success return. Every `?` and early
+            // `return` between the two now reaps, including ones added later by
+            // someone who never read this comment.
+            netns_guard = Some(net::NetnsGuard::new(&name));
 
             // Apply default-deny iptables policy BEFORE any process spawns.
             // This closes the race window where a process could exfiltrate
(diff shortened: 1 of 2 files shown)


brandon-coproduct:
Validated on the VM against this branch. **The guard holds on both paths.**

## Failed boots no longer leak

First run created 20 pods without tearing any down, which hits the known concurrency wall — **5 of 20 failed to boot**:

```
BEFORE  netns=5
        20 pods: ok=15 fail=5
AFTER   netns=5      <- five failed boots, zero new namespaces
        firecrackers=0  cgroups=0  taps=0
```

Before this change, a pod that died during boot left its namespace behind — that is how `nuc-f45b3cc8` and `nuc-59c6d719` were found. Five failures now produce zero.

## The sequential clause passes

Re-run as rubric iteration 8 actually specifies — one pod at a time, torn down before the next:

```
BEFORE  netns=5
        20 SEQUENTIAL pods: ok=20 fail=0
AFTER   netns=5  firecrackers=0  cgroups=0  taps=0
```

**20/20, no growth in any leak class.**

## One correction to my own testing

The first run was not the clause. "20 calls in sequence, each its own pod" means each is torn down before the next; my script created all 20 first, so by pod ~10 it was running ten concurrently on 4 vCPU and hitting the 30 s proxy-health timeout measured earlier at N=10. The 5 failures were a property of my script, not of the runtime, and I would have mis-recorded iteration 8 as failing had I stopped there.

State directories still grow (192 and counting). That is the retention question tracked separately — audit material with no rotation policy, not a leak to fix by deleting receipts.

PR #2409, opened 2026-09-03
perf(guest-init): time the vsock handshake, which was entirely untimed

brandon-coproduct:
A boot trace reported `unaccounted=334ms of 389ms` — **86% of guest startup unattributed**. Some of that is not merely unattributed, it is **outside the clock entirely**: `/init` performs the vsock handshake and *then* execs the tool-proxy, whose `nucleus-startup-trace` starts at its own process start. Everything `/init` did happened before any timer the host could read.

`nucleus-guest-init` had **zero** timing calls. It makes five sequential vsock round trips — identity, broker secret, mediation key, pod caller token, task token — and printed a line per fetch saying it happened, never how long it took.

## What this adds

Each leg is timed, plus a handshake total. Both are needed and they answer different questions: the **per-leg** times say *which* round trip is slow; the **total** says whether collapsing five into one is worth doing at all. Deciding to batch them without the total would be optimising a number nobody had.

`timed` mirrors the tool-proxy's helper, including printing per step rather than only in a summary — a summary that prints at the end cannot diagnose a hang, because the stall means the summary never runs and the console stays empty. Streaming makes the last line name the step that finished, so the stall is in the one after it.

The total prints even when no workload API port is configured, so a boot with no handshake reads as an explicit `0ms`. Otherwise "the line is missing" and "the handshake was free" look identical in a log — and the first is a bug while the second is a fact.

## Scope

Rubric **iteration 9** (close the instrumentation gap, target `unaccounted` under 15%) and the prerequisite for **iteration 12** (batch the handshake), which cannot be justified without knowing what it currently costs.

Printing only — no behaviour change. 11 tests pass; clippy and rustfmt clean. Measured numbers from a rebuilt guest to follow in a comment.

### DIFF
--- a/crates/nucleus-guest-init/src/main.rs
+++ b/crates/nucleus-guest-init/src/main.rs
@@ -49,6 +49,26 @@ const PROXY_BIN: &str = "/usr/local/bin/nucleus-tool-proxy";
 const EGRESS_PROBE_BIN: &str = "/usr/local/bin/nucleus-egress-probe";
 const GUEST_NET_SH: &str = "/usr/local/bin/guest-net.sh";
 
+/// Time one startup step and print it as it completes.
+///
+/// Mirrors `nucleus-startup-phase` from the tool-proxy deliberately, including
+/// printing per step rather than only in a summary: a summary that prints at the
+/// end cannot diagnose a hang, because a stall means the summary never runs and
+/// the console stays empty. Streaming means the LAST line names the step that
+/// finished, so the stall is in the one after it.
+///
+/// This half of the boot was entirely untimed. `/init` performs the vsock
+/// handshake and then execs the tool-proxy, whose own trace starts at ITS
+/// process start -- so everything here happened before any clock the host could
+/// read, and a boot trace reporting `unaccounted=334ms of 389ms` was not even
+/// measuring this part.
+fn timed<T>(name: &str, f: impl FnOnce() -> T) -> T {
+    let t = std::time::Instant::now();
+    let out = f();
+    eprintln!("nucleus-init-phase {name}={}ms", t.elapsed().as_millis());
+    out
+}
+
 fn main() {
     if let Err(err) = run() {
         eprintln!("nucleus-guest-init error: {err}");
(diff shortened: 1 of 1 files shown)


brandon-coproduct:
Measured on a rebuilt guest (aarch64 / KVM). **The numbers refute the reason I wrote this.**

```
nucleus-init-phase identity=20ms
nucleus-init-phase broker_secret=4ms
nucleus-init-phase mediation_key=2ms
nucleus-init-phase pod_caller_token=0ms
nucleus-init-phase task_token=0ms
nucleus-init-phase handshake_total=85ms

nucleus-startup-phase sandbox_proof=49ms  at=154ms
nucleus-startup-phase spec_read=8ms       at=172ms
nucleus-startup-phase audit_log=15ms      at=296ms
nucleus-startup-phase boot_report=5ms     at=428ms
nucleus-startup-phase vsock_bind=2ms      at=467ms
nucleus-startup-trace total=477ms unaccounted=398ms
```

## Batching the handshake is not worth doing

The five round trips sum to **26 ms** of a 477 ms boot — about **5%**. Rubric iteration 12 proposed collapsing them into one, on the assumption that five sequential vsock trips were a large share of startup. They are not, and the ceiling on that optimisation is 26 ms. **Iteration 12 should be dropped or deprioritised**, which is a better outcome than spending a day building the batched protocol and discovering this afterwards.

Two things worth noting inside the handshake itself: `handshake_total=85ms` against 26 ms of fetches means **59 ms is the work between them** — writing the SVID, exporting env, the processing after each response — not the transport. And `identity=20ms` is most of the transport cost, the other four being 4/2/0/0.

## Where the time actually is

The `at=` offsets localise the 398 ms unaccounted to four gaps *between* the tool-proxy's existing phases:

| gap | ~cost |
| --- | --- |
| process start → `sandbox_proof` | ~105 ms |
| `spec_read` → `audit_log` | ~109 ms |
| `audit_log` → `boot_report` | ~117 ms |
| `boot_report` → `vsock_bind` | ~37 ms |

That is the next instrumentation target, and it is a much better-posed one than "add spans somewhere" — each gap is bounded by two known lines.

The `unaccounted` figure did not drop, because this PR instruments a region that was never inside `total` to begin with. That is the finding, not a shortfall: 85 ms of boot was happening before the trace started.

PR #2410, opened 2026-09-03
perf(tool-proxy): checkpoints for the startup stretches spans could not reach

brandon-coproduct:
`nucleus-startup-trace total=477ms unaccounted=398ms` — **83% of tool-proxy startup unattributed**, despite six phases already being timed. The phases were not wrong, they were *sparse*. Reading the `at=` offsets localised the missing time to four stretches **between** them:

| stretch | ~cost |
| --- | --- |
| process start → `sandbox_proof` | ~105 ms |
| `spec_read` → `audit_log` | ~109 ms |
| `audit_log` → `boot_report` | ~117 ms |
| `boot_report` → `bind` | ~37 ms |

## Why `timed` could not cover them

It wraps a future. Those stretches are straight-line statements with `?` and `await` interleaved — rustls provider install, tracing-subscriber init, arg parsing, runtime and approver construction, attestation verifier, node client, credential loading. Wrapping them means restructuring code that has nothing to do with measurement, and the restructuring is where mistakes get made.

`mark` measures the stretch since the previous milestone instead, so attributing a region costs **one inserted line** and no change to control flow.

## The subtle part

`timed` now also resets the checkpoint clock. Without it, a `mark` following a timed phase would count that phase a second time — the milestones would stop partitioning the timeline and the parts could sum to more than the total. That is worse than no trace, because it looks precise.

## Scope

Rubric iteration 9 (target `unaccounted` under 15%). Measured effect to follow in a comment. What this changes regardless: the remaining unaccounted time is now bounded by named checkpoints rather than spread across the whole of `main`.

370 tests pass. Clippy clean under **both** `-p` and `--all-targets` — the flag difference that let a dead-code error through to CI on #2408.

PR #2411, opened 2026-09-03
perf(tool-proxy): close the last big gap, and split the biggest phase

brandon-coproduct:
Stacked on #2410 (which took `unaccounted` from 83% to 21%, against a rubric target of 15%).

Working the remaining 94 ms out of the `at=` offsets showed it was **not** spread evenly — one gap held 36 ms of it:

```
gap before sandbox_proof   11ms
gap before audit_log       16ms
gap before vsock_bind      36ms   <- the one worth a line
(five others, 5-10ms each)
```

## Why that gap existed

`serve_prep` sits in front of `TcpListener::bind`, and **a booted microVM never reaches it** — the guest serves over vsock and returns from the block above. So the router build, the watcher spawns and the layer stack went unmeasured on the only path guest boot takes. `router_build` is placed **before** the transport branch so it covers both.

Expected: unaccounted 94 ms → ~58 ms of 443 ms, about **13%**, which would meet the target. Measured numbers to follow in a comment rather than asserted here.

## Splitting the biggest phase

`logging_init` was 96 ms and named three things at once. It is now `crypto_provider` / `tracing_init` / `args_parse`, so whoever optimises it knows which one to attack.

One trap recorded at the site: **the rustls provider install is eager on purpose.** rustls [panics on first TLS use](https://github.com/rustls/rustls/issues/1938) when no provider is installed, so "defer it to save 96 ms" is a footgun, and the comment now says so before someone tries it.

370 tests pass; clippy clean under both `-p` and `--all-targets`.

### DIFF
--- a/crates/nucleus-tool-proxy/src/main.rs
+++ b/crates/nucleus-tool-proxy/src/main.rs
@@ -1406,6 +1406,9 @@ async fn main() -> Result<(), ApiError> {
     // Install rustls crypto provider before any TLS connections (web_fetch, etc.).
     let mut st = startup_trace::Startup::new();
     let _ = rustls::crypto::ring::default_provider().install_default();
+    // Eager on purpose -- rustls panics on first TLS use if no provider is
+    // installed, so this cannot simply be deferred to save boot time.
+    st.mark("crypto_provider");
 
     {
         use tracing_subscriber::prelude::*;
@@ -1424,9 +1427,10 @@ async fn main() -> Result<(), ApiError> {
             )
             .init();
     }
+    st.mark("tracing_init");
 
     let args = Args::parse();
-    st.mark("logging_init");
+    st.mark("args_parse");
 
     // === Auth-secret sanity (fail-closed on a world-known key) ===
     // `auth_secret` is a required arg, but an EMPTY string satisfies "present"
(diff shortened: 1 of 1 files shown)


brandon-coproduct:
Measured on a rebuilt guest. **The rubric target is met.**

| | unaccounted | of total | |
| --- | --- | --- | --- |
| before any of this | 398 ms | 477 ms | **83%** |
| after #2410 | 94 ms | 443 ms | **21%** |
| after this PR | **62 ms** | 458 ms | **13.5%** |

Rubric iteration 9 asks for under 15%.

```
crypto_provider=14ms  at=14ms
tracing_init=21ms     at=36ms
args_parse=68ms       at=104ms
sandbox_proof=35ms    at=145ms
spec_read=4ms         at=154ms
runtime_build=92ms    at=251ms
audit_log=12ms        at=278ms
state_build=109ms     at=393ms
boot_report=8ms       at=406ms
router_build=32ms     at=443ms
vsock_bind=1ms        at=448ms
```

`router_build` landed at **32 ms**, closely matching the 36 ms hole the arithmetic predicted — which is a decent check that the checkpoint went where I thought it did.

## The split corrected my assumption

I expected the 96 ms `logging_init` to be dominated by the rustls provider install. It is not:

- `crypto_provider` = **14 ms**
- `tracing_init` = **21 ms**
- `args_parse` = **68 ms** ← larger than the other two combined

So the thing I was tempted to optimise (and which the [rustls docs warn against deferring](https://github.com/rustls/rustls/issues/1938)) is the *smallest* of the three. Worth stating precisely: `args_parse` names the stretch between the `tracing_init` checkpoint and the one after `Args::parse()`, so 68 ms is that whole stretch rather than provably `clap` alone — narrowing it further is one more checkpoint, and now clearly worth spending.

## Where boot time actually goes

`state_build` 109 ms + `runtime_build` 92 ms + `args_parse` 68 ms = **269 ms of a 458 ms boot**. Those three are the Stage C targets. Nothing else exceeds 35 ms.

Total drifted 443 → 458 ms between runs (~3%), so treat single-run figures as ±3% and do not read the 15 ms as a regression from this PR.

PR #2412, opened 2026-09-03
docs(perf): record what stages A and B actually measured

brandon-coproduct:
The ledger said `TODO` for iterations **7, 8 and 9** while all three had been run and shipped, and the rubric still listed **iteration 12** as work worth doing after it had been refuted. A loop that does not record its results will redo them — the next tick would have started by re-deriving things already measured.

## What is now recorded

- **7 PASS** — an errno no longer masquerades as an authorization decision. Fourteen sites mapped every filesystem error to `PathDenied`, so a pod reported `access denied: path 'audit': Is a directory` with a 403 when nothing had denied anything (#2407). Could not be judged before #2401 made the reason visible.

- **8 PARTIAL** — 2 leaked netns, 0 cgroups, 0 taps, 189 state dirs, measured on a host with zero live Firecrackers. The two namespaces belonged to exactly the pods whose guests panicked, so cleanup ran on the arms somebody had remembered and not the late ones (#2408). The state dirs are audit material with no rotation policy — a **retention** question, deliberately not "fixed" by deleting audit data. The 20-call sequence is still to run.

- **9 PASS** — `unaccounted` 83% → **13.5%**, against a 15% target.

- **12 DROPPED.** The five vsock fetches cost **26 ms of a 458 ms boot**, so batching them has a 26 ms ceiling. **Struck through rather than deleted** — the reasoning that made it look worthwhile is the useful part, and a deleted row invites someone to propose it again next quarter.

## Also

The headline table's "11% accounted for" is now **86.5%**, with a pointer to the row that changed it.

And the three phases worth attacking are recorded where Stage C will look: `state_build` 109 ms, `runtime_build` 92 ms, `args_parse` 68 ms — **269 ms of a 458 ms boot**.

Documentation only.

### DIFF
--- a/docs/perf/RUBRIC-LEDGER.md
+++ b/docs/perf/RUBRIC-LEDGER.md
@@ -22,9 +22,9 @@ guest rootfs and harness must be built from one tree; skew presents as
 | 4 | TODO | | |
 | 5 | TODO | | |
 | 6 | TODO | | |
-| 7 | TODO | | |
-| 8 | TODO | | |
-| 9 | TODO | | |
+| 7 | PASS | 2026-09-03 | Refusals now carry a cause, and an errno no longer masquerades as one. Fourteen sites in `sandbox.rs` mapped EVERY filesystem error to `PathDenied`, so a live pod reported `access denied: path 'audit': Is a directory (os error 21)` with `kind=path_denied` and a 403 -- nothing had denied anything. Now classified by what the OS said: `PermissionDenied` -> `PathDenied` 403 (a real denial), `NotFound` -> 404, else 400. Same defect Linux hit from the other side (apparmor returned ENOENT when it denied). #2407, merged. Prerequisite was #2401, which made the reason visible at all -- before that this could not be judged. |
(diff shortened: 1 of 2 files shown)


brandon-coproduct:
Iteration 8 can move **PARTIAL → PASS** in a follow-up; recording the evidence here rather than pushing, since this PR is already queued.

Ran the remaining clause on the VM against #2408:

```
20 SEQUENTIAL pods (each torn down before the next): ok=20 fail=0
netns 5 -> 5   firecrackers=0   cgroups=0   taps=0
```

And separately, the failure path: a run that created 20 pods without teardown hit the known concurrency wall with **5 failures, and still produced 0 new namespaces** — which is the behaviour #2408 exists to give.

So row 8 becomes: 20/20 sequential, no growth in netns / cgroups / taps, with state-dir retention still the open item.

## Also worth sharpening in the rubric text

Iteration 8 reads "20 mixed calls in sequence, each its own pod". My first run took that as "create 20, each in its own pod" and created them all before tearing any down — which is a *concurrency* test, and it failed 5/20 for reasons that have nothing to do with the clause. The wording should say the pod is torn down before the next call starts, so the next person does not repeat that.

Both changes are small and belong in the same edit; deferring until this merges to avoid a conflict on the same two files.

PR #2415, opened 2026-09-03
fix(node): the health budget must scale with contention, not ignore it

brandon-coproduct:
`PROXY_HEALTH_TIMEOUT_SECS_DEFAULT` is **30 seconds regardless of how many microVMs are competing for the same cores**. A guest becomes healthy by running code, so on a 4-vCPU host N booting guests each get roughly 1/N of a core — time-to-ready grows with N while the budget does not.

## Measured on this hardware

| | outcome |
| --- | --- |
| N=1 | ready in ~3 s, comfortably inside 30 s |
| N=10 | **all ten failed** at the 30 s budget |
| N=10, budget raised | all ten became healthy, taking **~74 s** |

Nothing was wrong with those guests. They were slow because the host was busy, and a fixed deadline reported that as a pod failure — a system that works fine until you use it.

## The change

The budget scales with the number of live microVMs, **capped at 8×**. The cap is the part that matters: without it a host with fifty pods would grant each new one a 25-minute budget, so a genuinely broken guest would hang the caller instead of failing. The point is to stop punishing slow-because-busy, not to wait forever.

An explicit `NUCLEUS_NODE_PROXY_HEALTH_TIMEOUT_SECS` is still honoured verbatim — an operator who names a number is not asking to have it scaled behind their back.

## Testing

`health_budget_secs` is pure and total, so the policy is testable without booting anything. Five tests, two of which exist to catch how this goes *wrong* rather than how it goes right:

- the budget must be **monotonic** in contention — an arithmetic slip inverts that silently
- a pathological base must **saturate rather than wrap**

## Scope

This is rubric iteration 13's *problem*. It is not the whole of iteration 13, which asks for **edge-triggered readiness** — a protocol change worth doing on its own evidence. This removes the measured failure without one. [Adaptive timeouts that account for contention](https://www.researchgate.net/publication/401262495_ADAPTIVE_TIMEOUT_STRATEGIES_FOR_EFFICIENT_TRANSACTION_MANAGEMENT_UNDER_SNAPSHOT_ISOLATION_IN_DISTRIBUTED_DATABASES) are the standard answer to a fixed deadline under variable load.

395 tests pass; clippy clean under both flag sets; line ratchet passes.


brandon-coproduct:
Validated on the VM with **no explicit override** — the adaptive path is what ran:

```
override present? 0
```

## The fix (retest)

| N | before | after |
| --- | --- | --- |
| 10 | **0/10 failed at the 30 s budget** | **10/10, 74.6 s** |

That is rubric iteration 13's stated clause: *"N=10 passes with the stock timeout."*

## Everything around the fix (regression)

| N | before | after |
| --- | --- | --- |
| 1 | 3,011 ms · 1/1 | **3,013 ms · 1/1** |
| 5 | 10,902 ms · 5/5 | **10,384 ms · 5/5** |

The common case is unchanged — 2 ms apart at N=1, and N=5 marginally faster, both inside the ±3% run-to-run noise measured earlier. That matters because the obvious failure mode of "scale the timeout with load" is that it slows down or destabilises the uncontended path, and it does not.

The 74.6 s at N=10 matches the ~74 s measured when the budget was raised by hand, which is the expected result: the guests were always capable of becoming healthy, and the fixed deadline was the only thing calling them failures.

## What this does not claim

`largest clean burst` is now 10, up from 5. It is **not** a claim that 100 pods work — the ceiling has moved, not disappeared, and N=25/50 remain unmeasured against this change. Nor is it the whole of iteration 13, which asks for edge-triggered readiness; this removes the measured failure without a protocol change.

PR #2454, opened 2026-09-03
Move B: delete HMAC where mTLS replaces it, keep it where verification shows it's still load-bearing

brandon-coproduct:
## Move B — delete HMAC where mTLS actually replaces it

Part 2 of the mTLS migration (`sparkling-doodling-horizon.md`), landing as
one PR per the explicit "let's do it all in one PR" instruction. Move A
(mTLS between CLI, node, and tool-proxy↔node) is merged; this PR deletes
the HMAC auth tier **where verification showed it is genuinely dead**, and
explicitly does **not** delete it where verification showed it is still
load-bearing. That distinction is the actual output of this PR, not a
footnote — the original plan assumed a fully categorical deletion, and
direct verification (booting the reasoning, not just reading the code)
found that assumption wrong in three places. Sections below say exactly
which, and why.

### What's deleted, and verified safe

- **nucleus-node's own HTTP+gRPC HMAC tier** — `AuthMethod::Hmac`,
  `verify_http`/`verify_grpc`, `AuthConfig`, `ReplayCache`,
  `NUCLEUS_NODE_AUTH_SECRET`, and the `--http-mtls-self-issued`/
  `--grpc-tls-self-issued` flags that used to gate mTLS as opt-in. mTLS is
  now the **unconditional** default: `identity_manager` is always
  constructed, `http_serve::serve` always binds the mTLS listener,
  `serve_grpc` refuses to start without a client CA configured. Safe
  because the only two real callers — the CLI and tool-proxy's
  `node_client`/`lockdown_client` — are both converted below.
- **`nucleus-tool-proxy`'s `node_client.rs` and `lockdown_client.rs`** —
  the two internal clients that talk to the node — converted from HMAC
  signing to mTLS (reqwest for HTTP, a promoted `SpiffeServerCertVerifier`
  + tonic's `tls_config_with_verifier` for gRPC — this codebase's first
  tonic mTLS client).
- **`nucleus-tool-proxy`'s legacy HMAC approval tier**
  (`AuthTier::ApprovalHmacDrand`) — `/v1/approve` now accepts only an
  Ed25519 approver signature; a pod with no approver pubkeys configured is
  fatal at startup. Safe because the production (Firecracker) driver
  already used Ed25519 exclusively; only the local/container drivers'
  approval flow is affected, an accepted and documented trade.
- Four CLI callers found by grepping every `sign_http_headers` call site
  end to end, not by trusting that Move A's own pass had found them all:
  `nucleus node` (already had mTLS support from Move A step 5, now
  auto-defaults to the identity `nucleus setup` provisions instead of
  requiring explicit flags), `nucleus verify --tier2` (the tool the
  migration plan itself names for its own verification — was broken, so
  the plan could not have verified itself), the 2-safety harness
  (`twosafety_boot.rs`), and `nucleus run`'s remote-node path (likely the
  most-used of the four, and was hard-requiring an HMAC secret to even
  start).
- A new categorical denylist gate, `scripts/check-no-hmac-auth.sh`,
  scoped to exactly what's retired above — wired into CI and registered
  in `check-gates-can-fail.sh` with a real perturbation (verified RED on
  a reintroduced symbol, GREEN on restore).

### What's deliberately kept, and why — the actual finding of this PR

- **`nucleus-tool-proxy`'s residual `AuthTier::Hmac`**
  (`NUCLEUS_TOOL_PROXY_AUTH_SECRET`). Verified before touching it: this
  authenticates same-guest loopback callers — an in-guest `workload:`
  child process, and, on the local/container drivers (no vsock at all),
  the *primary agent's own traffic*, relayed through nucleus-node's
  `SignedProxy`. `AuthTier::HostVsock` does **not** cover this case:
  `pod_mgmt::peer_is_host` accepts only `VMADDR_CID_HOST`, and a
  same-guest loopback caller arrives as `VMADDR_CID_LOCAL` — a
  deliberately weaker, different guarantee. Deleting this without a
  replacement (per-workload SVIDs, or mTLS over vsock — an explicit gap
  the original plan already calls out as future work) would have broken
  every Firecracker pod that configures a `workload:`, not an edge case.
- **`nucleus-tool-proxy`'s sandbox-proof Tier 3** (`OrchestratorToken`).
  An earlier pass in this session tentatively planned to delete this too;
  verification found it's the *only* sandbox proof `nucleus run --local`
  and `nucleus shell` have — Tier 1/2 (SVID-based) don't apply to a bare
  local subprocess with no VM/container isolation. Left untouched.
- The workspace `Cargo.toml` version bump to 3.0.0 (discussed earlier in
  this work) is **not** in this PR: dozens of internal crates pin sibling
  dependencies to a literal `version = "1.0.0"` instead of
  `version.workspace = true`, so bumping the workspace version breaks
  resolution across the tree. That's a real, separate mechanical cleanup
  this PR chose not to bundle in.
- `nucleus-spec::tier2_artifacts`'s guest-release floor is unchanged — it
  tracks the repo's release TAG (currently v2.2.0), which this PR does not
  cut.

### Known, deliberately deferred gap

- `nucleus-perf` (the `podburst` load-testing tool) still signs its node
  requests with HMAC and defaults to `http://`. It will fail loudly
  (connection/protocol error) against a real node — an acceptable failure
  mode for a benchmarking tool, but a real follow-up.

### Verification

Per crate (`nucleus-node`, `nucleus-tool-proxy`, `nucleus-cli`,
`nucleus-identity`, `nucleus-guest-init`): `cargo test`, `cargo clippy`
under both `-p` and `--all-targets`, `cargo fmt --check`. Full workspace
`cargo check --workspace --exclude nucleus-verifier-service` (that one
crate has a pre-existing, unrelated missing-WASM-artifact gap).
`scripts/check-line-ratchet.sh --strict`. `scripts/check-no-hmac-auth.sh`.
A manual vendor-neutrality grep. `scripts/check-gates-can-fail.sh` (all 16
probed gates RED on their own subject, GREEN on restore, including the
new one).

Several of the new/changed code paths ship a **real** mTLS-handshake test
(not just construction-succeeds) — a genuine server, a genuine client
cert, a genuine refusal from an unrelated CA where that's the property
under test:
`nucleus-tool-proxy/src/node_client.rs::mtls_tests`,
`nucleus-tool-proxy/src/lockdown_client.rs::mtls_tests` (a full tonic
gRPC round trip, streaming a real command through),
`nucleus-node/src/auth.rs`'s SPIFFE-branch refusal test,
`nucleus-cli/src/node.rs`'s pre-existing pair, and a new one for
`nucleus-cli/src/run.rs::create_pod_via_node`.


bcrisp:
Merged \`origin/main\` in to clear the stale \`audit\` failure (the \`cargo-audit\`/tinyvec CI flake fixed on main earlier today — unrelated to this PR's own content).

One genuine, content-specific CI break remains: **"The node records pod lineage on the live path (C2)"** fails with \`error: unexpected argument '--auth-secret' found\` when \`scripts/cross-pod-lineage-check.sh\`/\`cross-pod-scoped-check.sh\` start \`nucleus-node\`. Root cause: these two scripts are the only ones in the repo that boot a bare local \`nucleus-node\` + CLI pair directly (no VM, no \`Tier2Host\`), and they still pass the now-retired \`--auth-secret\` flag Move B removed from the node's own CLI surface.

I looked for the right fix and didn't find one I trust — flagging instead of guessing:

- The only mTLS-identity-provisioning code that exists (\`provision::provision_mtls_identity\`/\`mint_cli_identity\`) is built around a \`Tier2Host\` (SSH-provisioned real/VM host) or a pod's node-minted identity. Neither fits a bare local test process.
- \`scripts/check-no-hmac-auth.sh\`'s own denylist explicitly retires \`--http-mtls-self-issued\`/\`--grpc-tls-self-issued\`/\`NUCLEUS_NODE_HTTP_MTLS_SELF_ISSUED\`/\`NUCLEUS_NODE_GRPC_TLS_SELF_ISSUED\` — i.e. the "node self-issues a cert and a script captures it" pattern that would have made this tractable was deliberately removed, not just relocated.
- So a bare local node process + a separately-invoked CLI process currently have no way to establish a shared trust root in this test harness's shape. That's either a real gap in the two scripts' rewrite, or a sign these two checks need a different harness shape post-Move-B (e.g. driven through the same in-process path \`twosafety_boot.rs\`/\`nucleus verify\` already use, rather than raw shelled-out binaries).

Left auto-merge armed (harmless while blocked) rather than guessing at nucleus's own mTLS bootstrap design.

brandon-coproduct:
Update: fixed for real, not just diagnosed.

The earlier comment on this thread identified that `cross-pod-lineage-check.sh`/`cross-pod-scoped-check.sh` were passing the retired `--auth-secret` flag. Digging further surfaced a second, independent, genuine production bug behind it: `nucleus-node/src/main.rs`'s local-driver `enable_pod_mgmt` pod-spawn code never got updated when Move B made the node's HTTP API mTLS-only — it was still pointing the orchestrator pod's tool-proxy at `http://` and passing a dead `NUCLEUS_TOOL_PROXY_NODE_AUTH_SECRET` env var the receiving side had already stopped reading. Any `enable_pod_mgmt: true` pod on this branch would fail at tool-proxy startup (`--enable-pod-mgmt requires an SVID identity ... to authenticate to nucleus-node over mTLS`) — not a test-only gap.

Fixed in b01ea713:
- The node now mints the orchestrator pod its own SVID via the already-running `IdentityManager` (same mint-and-cache path pod certificates already take), writes cert/key/trust-bundle under the pod's own state dir, and hands them to the spawned tool-proxy via `NUCLEUS_IDENTITY_CERT`/`_KEY`/`_TRUST_BUNDLE` — env-parity with what guest-init already does on the real Firecracker path over vsock.
- Both C2 scripts switched to `https://` and given a lightweight way to bootstrap a client identity for a bare local node+CLI pair (no VM/Tier2Host in the loop): a new `nucleus-identity` example, `mint_test_client_cert`, that loads the node's own already-persisted local CA and mints a client cert signed by it.

Verified end-to-end against freshly rebuilt binaries, not just compiled — both scripts pass for real now. Full sweep also green: `cargo test -p nucleus-node --features local-driver` (395 passed), `-p nucleus-identity -p nucleus-tool-proxy` (all green), clippy `--all-targets -D warnings` on both touched crates, fmt, `ci/no-vendor-strings.sh`, `scripts/check-no-hmac-auth.sh`, `scripts/check-line-ratchet.sh`.

Re-queued for merge.

Issue #2558, opened 2026-09-04
[epic] OS/Firecracker boundary as formal objects: typed lifecycle, typestates, refusals, named assumptions


brandon-coproduct:
## Problem
The Linux/Firecracker isolation surface (jailer, seccomp, netns, mounts, rlimits, vsock identity, guest-init ordering) is entirely test-enforced or unenforced — zero Kani in `nucleus-node`/`nucleus-tool-proxy`/`nucleus-guest-init`; the only formal objects touching it are the egress-chain model and the FM-5 env classifier. Live defects: six `?` early returns in `spawn_firecracker_pod` bypass teardown and the VMM `Command` has no `kill_on_drop`; host-namespace iptables rules are never removed; `mount_fs` ignores errors; off-root spawn silently skips uid-drop/`NO_NEW_PRIVS`/rlimits; `SeccompSpec::Disabled`/`Custom` and `--firecracker-jailer=false` are reachable in production; `admit`-before-spawn is enforced by an `include_str!` source-text test; vsock peer-CID unforgeability is prose in three comments.

## Decision (owner-approved 2026-09-04)
Typed lifecycle (`PodResources` linear must-use handle + fault-injection matrix, since `Drop` cannot await), guest-init typestate, `AdmittedPodPlan` token, production refusals for all three weak knobs, jailer `--new-pid-ns`, rlimit lattice ≤ policy, named kernel-behaviour assumptions in FM-5's not-claimed list (never a citable theorem), a Flux refinement-type spike on the pure rlimit module (nightly, non-required), requireable boot-gated probes. seccompiler BPF hash-pinning is deferred behind the refusals.

## Governing rules (apply to every sub-issue)
1. A gate that can pass without measuring is deleted or fixed before anything depends on it; every gate gets a `scripts/check-gates-can-fail.sh` probe that plants a regression and asserts red.
2. Nothing is advertised that is not run: every `#[kani::proof]` is named by a job or on a shrinking allowlist; every `lean_lib` maps to a concrete workflow step; documentation numbers are generated, never typed.
3. Two independent kernels check the proven tier; axiom sets are asserted per theorem; `implemented_by`/`extern`/`csimp` are banned in proven sources and generated output.
4. Un-extractable OS code gets typestates + fault-injection tests, never a fake Lean/Kani artifact.
5. Vendor neutrality (CLAUDE.md): external proof-search engines are named generically.

## Scope
- [ ] #2579
- [ ] #2588
- [x] #2570
[...]

PR #2631, opened 2026-09-05
security(guest-init): boot typestate — exec only from Sealed, load-bearing mounts fail closed, no shell fallback (#2589)

### LINKED ISSUE #2589: guest-init typestate (remount-before-exec as a type), fail-closed load-bearing mounts
Part of #2558 · Wave 1 · Size L · Blocked by: none

## Problem
`nucleus-guest-init/src/main.rs::run` fetches every secret over vsock (`:146-330`) before `remount_root_ro` (`:471`) before `exec_proxy` (`:494`) — correct ordering enforced only by source order. `mount_fs` (`:594-608`) ignores every error; `resolve_pod_spec` (`:659`) drops to `/bin/sh` as PID 1 on a missing spec; `main()` (`:72-76`) swallows `run()`'s error (fail-closed only by kernel panic on PID-1 exit).

## Fix
Typestate `Booting→Mounted→Provisioned→Sealed→Executing` so `exec` requires `Sealed` (remount-before-exec is a type error otherwise); classify `GUEST_MOUNTS` entries load-bearing vs optional (read-only `/work` is legitimate per `workload.rs:392-400`) and make load-bearing mounts fail-closed; refuse the `/bin/sh` fallback; propagate `run()` errors.

## Acceptance criteria
- Compile-time: no path reaches `exec_proxy` without `Sealed`; a failed load-bearing mount aborts boot with a named error.

brandon-coproduct:
Closes #2589 (part of #2558, Wave 1).

## What
- **`nucleus_guest_init::boot`** (new lib target beside the bin): `Boot<Booting> → mount_all → Boot<Mounted> → provisioned → Boot<Provisioned> → seal → Boot<Sealed> → exec`. `seal` *is* the read-only remount and the only way to obtain `Boot<Sealed>`; `exec` exists only there and mints a `SealedProof` (no public constructor) that the launcher `exec_proxy` demands. Reordering remount and exec, skipping the mounts, or forging the proof are **type errors** — three `compile_fail` doctests pin that, and a fourth doctest shows the launcher runs only after the remount.
- **Load-bearing vs optional mounts**: `GuestMount` carries `load_bearing` (all five pseudo-filesystems today); `mount_fs` returns its error instead of printing and continuing; a failed load-bearing mount aborts the boot with `BootError::LoadBearingMountFailed { target, fstype, error }`. `/work` (`/dev/vdb`) stays optional — a read-only or absent data volume is a legitimate guest — and optional failures are recorded on the boot.
- **No `/bin/sh` fallback**: `resolve_pod_spec` returns `BootError::PodSpecMissing` naming both paths ("a shell as PID 1 would be an unmediated workload").
- **`run()` errors propagate**: `main` prints the error and exits 1, so the kernel panic on PID 1 exit is the deliberate fail-closed outcome with its reason on the console, not the by-product of swallowing an error.

## Acceptance
- Compile-time: no path reaches `exec_proxy` without `Sealed` (the launcher's `SealedProof` parameter; doctests).
- A failed load-bearing mount aborts boot with a named error (unit tests over an injected `mount(2)`).

Tests: 6 boot unit tests, 11 existing tests, 4 doctests (3 `compile_fail`); clippy clean.

https://claude.ai/code/session_01GZaH5waz88vL1qfNpFNfZ4

### DIFF
--- a/crates/nucleus-guest-init/Cargo.toml
+++ b/crates/nucleus-guest-init/Cargo.toml
@@ -8,6 +8,10 @@ description = "Guest init for Firecracker images"
 repository = "https://github.com/coproduct-opensource/nucleus"
 readme = "README.md"
 
+[lib]
+name = "nucleus_guest_init"
+path = "src/lib.rs"
+
 [[bin]]
 name = "nucleus-guest-init"
 path = "src/main.rs"
--- a/crates/nucleus-guest-init/src/boot.rs
+++ b/crates/nucleus-guest-init/src/boot.rs
@@ -1,0 +1,336 @@
+//! The guest boot as a typestate: `Booting → Mounted → Provisioned → Sealed`,
+//! and `exec` only from `Sealed` (#2589).
+//!
+//! `run()` used to fetch every secret over vsock, then `remount_root_ro`, then
+//! `exec_proxy`, with the ordering enforced by nothing but the order of the
+//! lines. Reorder them — move the remount after the exec, or forget it — and
+//! the workload starts on a writable rootfs with no compiler, test or gate
+//! noticing. Here the order is the type: [`Boot<Sealed>`] is the only state
+//! with an `exec`, it is only produced by [`Boot::<Provisioned>::seal`], which
+//! is the remount, and `Provisioned` is only produced from `Mounted`. A boot
+//! that skips a step does not compile (see the `compile_fail` doctests in
+//! `lib.rs`).
+//!
+//! Two failure policies live here too, because they are the same kind of
+//! claim ("boot does not proceed on a broken base"):
+//!   * mounts are classified **load-bearing** or optional; a load-bearing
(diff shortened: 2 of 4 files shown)


brandon-coproduct:
The x86_64 boot lane red was real, and this PR's own doing: the guest kernel auto-mounts devtmpfs, so the now-load-bearing `/dev` mount returned EBUSY and the typestate aborted the boot (`nucleus-guest-init error: load-bearing mount /dev (devtmpfs) failed … EBUSY — boot aborted`, then the kernel panicked on init exit). The old fail-open `mount_fs` had hidden this behind a log line for as long as it existed.

Fix in 33092ab4: on EBUSY, require the target to actually be a mount point (pure stat, works before /proc) and then remount with the list's flags, since the kernel's automount carries none of them and a `/dev` without nosuid is not the `/dev` the mount table promises. Any other error still aborts.

Issue #1636, opened 2026-05-26
[LOW-1] Derive executor_id from persistent pubkey (post-HIGH-1)


brandon-coproduct:
**Tier 2 / LOW-1.** Part of #1626. **Depends on HIGH-1 (#1630)**. See [roadmap #1624](https://github.com/coproduct-opensource/nucleus/issues/1624).

## Problem

`crates/nucleus-node/src/trust_gate.rs:71-83` — `executor_id` derives from `uuid_hex(pid, nanos)`. Predictable and not collision-resistant under adversarial clock control. Currently masked by the bigger HIGH-1 bug (key regen on restart); once HIGH-1 lands, this becomes the meaningful stability bug.

## Approach

After HIGH-1 (#1630) lands, change:

```rust
executor_id = hex::encode(&sha256(verifying_key_bytes)[..8])
```

Stable across restarts, collision-resistant, ties identity to the signing material.

## Files

- `crates/nucleus-node/src/trust_gate.rs:71-83`

## Acceptance

`executor_id` stable across two restarts (uses same persisted key from HIGH-1).

Issue #1635, opened 2026-05-26
[MED-1] Rename nucleus-lineage 'dev' feature to 'insecure-local-issuer'


brandon-coproduct:
**Tier 2 / MED-1.** Part of #1626. See [roadmap #1624](https://github.com/coproduct-opensource/nucleus/issues/1624).

## Problem

`crates/nucleus-lineage/Cargo.toml:18` — `dev = [...]` is ergonomically named and might get enabled "for deterministic tests" and shipped. The cargo-boundary gating is airtight today; the risk is downstream-consumer footgun at the `[dependencies]` line.

## Approach

- Rename `dev = [...]` → `insecure-local-issuer = [...]`.
- Add transitional alias `dev = ["insecure-local-issuer"]` for one release with a doc-comment + CHANGELOG entry (Cargo doesn't support `#[deprecated]` on features formally).
- Update internal callers (examples, tests) to use new name.

## Files

- `crates/nucleus-lineage/Cargo.toml:14-18`
- `crates/nucleus-lineage/src/lib.rs:26+` — docstring
- Any `--features dev` in CI/Makefile/scripts (grep workspace-wide)

## Acceptance

Workspace builds with `--features insecure-local-issuer` AND with `--features dev` (transitional alias).

Issue #1634, opened 2026-05-26
[CRIT-2] Wire CedarEvaluator into kernel decision path; flip cedar feature default-on


brandon-coproduct:
**Tier 2 / CRIT-2.** Part of #1626. See [roadmap #1624](https://github.com/coproduct-opensource/nucleus/issues/1624).

## Problem

`crates/portcullis/src/cedar_bridge.rs:53-243` defines a complete `CedarEvaluator` but only its own tests call it. The `cedar` feature flag at `crates/portcullis/Cargo.toml:34` is default-off. Buyer-facing docs claim Cedar policy enforcement.

## Approach — wire Cedar as an optional policy gate in `Kernel`, default-on

- **Field**: `Kernel.cedar_evaluator: Option<CedarEvaluator>` (constructor accepts).
- **Hook**: in `decide_term()` at `crates/portcullis/src/kernel.rs:1534+`, after preflight + after FlowTracker check (CRIT-1 #1632), invoke Cedar if present:
  - Map (`session_id`, `Operation`, `subject_path`, `IFCLabel.integrity`, `.authority`, `.confidentiality`) → Cedar context per `cedar_bridge::evaluate()` at `crates/portcullis/src/cedar_bridge.rs:91-170`.
  - `cedar_policy::Decision::Deny` → `Verdict::Deny(DenyReason::CedarPolicyDenied)`.
- **Loading**: convention `state_dir/policies/*.cedar`. Load at startup; reload on SIGHUP.
- **Feature default**: flip `cedar` default-on in `crates/portcullis/Cargo.toml:34`; re-export feature in `crates/nucleus-tool-proxy`.

## Files

- `crates/portcullis/src/kernel.rs:420, 547, 1534-1542` — field, ctor param, decision hook
- `crates/portcullis/Cargo.toml:34` — flip `cedar` default-on
- `crates/portcullis/src/cedar_bridge.rs` — add `from_directory(path)` loader

## Reference

[Cedar joins CNCF Sandbox (Jan 2026) — InfoQ](https://www.infoq.com/news/2026/01/cedar-joins-cncf-sandbox/) — Cedar is the de-facto AI-agent policy language. [Bedrock AgentCore Policy + Cedar](https://medium.com/@tolghn/amazon-bedrock-agentcore-policy-cedar-control-what-your-ai-agents-can-do-3313e6f3a2db) — same architecture choice as Nucleus.

## Acceptance

[...]

Issue #1633, opened 2026-05-26
[CRIT-1] Wire FlowTracker into Rust MCP runtime — match Python SDK behavior


brandon-coproduct:
**Tier 2 / CRIT-1.** Part of #1626. See [roadmap #1624](https://github.com/coproduct-opensource/nucleus/issues/1624).

## Problem

The pitch claims "the only IFC library shipping in production." The Python SDK (`crates/portcullis-effects/src/runtime.rs:32+`) wires `FlowTracker` — the Rust runtime does not. Six tool-call entry points (`read`, `write`, `run`, `glob`, `grep`, `web_fetch` at `crates/nucleus-tool-proxy/src/mcp.rs:227-954`) call `kernel_decide()` but never observe `NodeKind` nodes or check `SafetyCheck`.

## Approach — mirror the Python SDK's `observe → check → execute` rhythm

- **Reuse** `FlowTracker` API at `crates/portcullis-core/src/ifc_api.rs` (`observe`, `observe_with_parents`, `check_action_safety`, `check_exfiltration_safety`).
- **Add field** `NucleusMcpServer.flow_tracker: Arc<Mutex<FlowTracker>>` at `crates/nucleus-tool-proxy/src/mcp.rs:118-130`.
- **Per entry-point**:
  - `read` (L227): `observe(NodeKind::FileRead)` → kernel → safety check before sandbox call
  - `write` (L291): `check_exfiltration_safety(parent_node_id, sink_max_conf)` before write
  - `run` (L361): `check_action_safety(node_id, requires_authority=true)` — shell exec needs authority
  - `glob`/`grep` (L446, L569): `observe(NodeKind::FileRead)` + safety check
  - `web_fetch` (L757-954): `observe(NodeKind::WebContent)` *after* successful fetch (response is adversarial-integrity); return node_id to caller so downstream operations track ancestry
- **Kernel hook**: extend `Kernel::decide_term()` at `crates/portcullis/src/kernel.rs:1534+` to accept `Option<&FlowTracker>` and consult `check_action_safety` before `policy_rules.evaluate()` at line 1278. `SafetyCheck::Reject` → `Verdict::Deny(DenyReason::IfcUnsafe)`.
- **Session ceiling propagation**: ensure `session_taint_ceiling()` and `session_conf_ceiling()` are read on each decision.

## Files

- `crates/nucleus-tool-proxy/src/mcp.rs:118-130` — add `flow_tracker` field
[...]

PR #2716, opened 2026-09-09
Content-addressed pod identity and a snapshottable launch path

brandon-coproduct:
Three commits, readable in order. Each stands alone; the third is where the substrate work is.

## Read hardest

**`image_identity.rs` — the TOCTOU decision.** `admit_posture` measures a rootfs ~1300 lines before `prepare_jail` places it, and anything that swaps the file in between is measured as one thing and booted as another. The obvious fix was `linkat(AT_EMPTY_PATH)` from the measured fd, which wants `CAP_DAC_READ_SEARCH`. This verifies **after placement** instead: `place_resource` hard-links into the jail, so the in-jail path is the same inode, and where it falls back to a copy, the copy is what boots. No capability, no window.
*The question to ask:* is post-placement verification right for the two non-jailed branches, where nothing is placed and the pre-existing window remains? It is named in the module doc rather than papered over.

**`snapshot.rs` / `workload_api_*` — who answers what.** A snapshot base needs two facts: far enough along to be useful, not far enough along to be somebody. The guest declares the first (`SNAPSHOT_READY`); the host decides the second from its own record of what it served. Neither is trusted for the other's question, because the guest is the thing being contained.
*The question to ask:* is `personalizes_the_vm()`'s classification right? It is exhaustive, so a new command is a compile error — but the split itself is a judgement.

## Measured, not assumed

Validated on real KVM (Firecracker v1.16.1, the pinned version, under Lima):

| | measured | previously documented |
|---|---|---|
| snapshot restore + resume | **10 ms** | 7 ms — holds |
| snapshot create | **480 ms** | 151 ms — ~3x optimistic |

Both boot modes reach `CREATE 200` with the full vsock credential chain; a snapshot round-trips into a second VMM.

**One design claim measured false.** The API socket was expected to buy verify-then-boot — a VMM idling with its seccomp filter installed, checkable while stopped. It does not: Firecracker installs the filter at `InstanceStart`, not at exec, and a VMM left idle five seconds still reports `seccomp mode 0`. The check stays downstream exactly as for a config file. The reasoning is recorded at the call site so nobody re-derives the wrong order.

## Safety of the change

- **API-mode boot is default off** (`--firecracker-boot-mode`). The config-file path is unchanged and was regression-tested end to end.
- **Absent image digests are a no-op**, so every existing spec is unaffected.
- `SNAPSHOT_READY` is **optional**, so no `GUEST_RELEASE_FLOOR` bump: a guest that never sends it is refused as a base, which is the right answer for an image that cannot say where its barrier is.

## What I would question first

`main.rs` grew to its ratchet ceiling and the ratchet was raised three times, each with a written reason. The growth is the kind the ratchet wants — ~400 lines went into new, separately testable modules — but a module declaration cannot be extracted, so it is paid for in `main.rs`. Worth deciding whether that trade is acceptable or whether the ceiling should hold and something else move out.

Separately: `fix(boot-harness)` restores a script that `docs/production-delta.md` lists as Done and "Verified by running it from clean: CREATE 200". It had not run since Move B deleted the HMAC auth it depends on, and nothing noticed because no workflow runs it.

---

### Velocity panel

**https://claude.ai/code/artifact/cac5cf92-bf88-40fd-83e2-801cb188bab9**

A small dashboard for the numbers this branch produced. Every figure carries its own provenance —
**measured** (observed on real hardware), **declared** (a configured value), or **projected**
(arithmetic on those two) — because that is the argument this system makes about receipts and it
should hold for its own dashboard.

The headline is 79 ms cold boot against 17 ms restore-and-resume, measured on Firecracker 1.16.1
with vsock and a network interface attached. The panel also states the unflattering reading, which
is the useful one: **62 ms saved against a 2 400 s gate is noise.** What the substrate buys is not
a faster cold run but a reusable one, and the number that would show it — a measured cold wall time
for `clippy` and `test-core` — does not exist yet. ADR 0003 requires it before either gate moves.

https://claude.ai/code/session_01BnBVBxkRDqrRUNbjxPqS1s


brandon-coproduct:
Shepherding note — this is why auto-merge is armed but the PR never enqueues.

`Code Coverage (llvm-cov)` fails on **step 7, `Workspace coverage (threshold gate)`**, not on a test. Every test in the run passes. The gate is `--fail-under-lines 83` (`.github/workflows/coverage-matrix.yml:148`) and this branch reports:

```
TOTAL   241962  40160  83.40%   16912  3304  80.46%   156483  26765  82.90%
                       regions                functions              lines
```

82.90% against a floor of 83 — short by a tenth of a point. `main`'s merge groups clear it (last five `merge_group` coverage runs are all green), so this is the branch's own delta, not a moved floor.

The cheapest tenth of a point is not in this PR's new code. Three `xtask` files carry 752 uncovered lines between them at **0.00%**:

| file | uncovered lines |
|---|---|
| `xtask/src/ci_timings.rs` | 270 |
| `xtask/src/ci_otel.rs` | 264 |
| `xtask/src/ci_spec.rs` | 218 |

Covering any one of them moves the workspace number by more than the gap. That is a separate PR, though — the honest fix *here* is tests for the content-addressed identity and snapshot-launch paths this PR adds, since those are what pulled the average down.

Also red and separable: `Build (macOS, release crates)` hit the 35m0s wall (timeout, not a compile error), and `CodeQL` failed in 2s (infrastructure). Both want a re-run rather than a code change.

brandon-coproduct:
Correction to my note above: **CodeQL is not infrastructure.** It failed in 2s because it had nothing to build — it reported on an already-finished analysis — and what it reported is nine *new* high-severity alerts, all in this PR's own new file:

| alert | rule | location |
|---|---|---|
| #1628 | `rust/path-injection` | `crates/nucleus-node/src/snapshot_store.rs:298` |
| #1629 | `rust/path-injection` | `snapshot_store.rs:453` |
| #1630 | `rust/path-injection` | `snapshot_store.rs:395` |
| #1631 | `rust/path-injection` | `snapshot_store.rs:393` |
| #1632 | `rust/path-injection` | `snapshot_store.rs:327` |
| #1633 | `rust/path-injection` | `snapshot_store.rs:435` |
| #1634, #1635 | `rust/path-injection` | `snapshot_store.rs:315` |
| #1636 | `rust/path-injection` | `snapshot_store.rs:400` |

All nine trace to the same sink, `published_dir`:

```rust
pub fn published_dir(&self, derivation: &Derivation) -> PathBuf {
    self.root.join("by-derivation").join(derivation.name())
}
```

I believe these are false positives, and the doc comment above that function already argues why — it takes the `Derivation` rather than its name precisely so no `pub fn` here joins a caller-supplied string under the store root. Checking the argument rather than trusting it: `Derivation::name()` (`snapshot_store.rs:126`) is `hex::encode(Sha256::finalize(..))` over a canonicalized serialization. Sixty-four characters from `[0-9a-f]`. It cannot contain a separator, cannot be absolute, cannot contain `..`. The escape `Path::join` allows is unreachable.

CodeQL's dataflow does not know that, and `hex::encode` is not one of its recognized sanitizers — the taint runs from the `Serialize`d public fields of `Derivation` straight through the digest to the join.

So the fix is one of two, and it is a real decision:

1. **Dismiss the nine as `false positive`** with that reasoning, matching how the Scorecard batch was handled. Fastest, and the structural argument is already written down in the source.
2. **Give CodeQL a barrier it recognizes** — have `name()` return a newtype whose constructor rejects anything not 64 hex characters, so the guarantee is a runtime check on the path rather than an inference about SHA-256. Slower, and arguably redundant, but it makes the check honest for the next person who adds a sink here.

Either way this is a separate axis from the coverage gap in my previous comment — the PR needs both cleared before auto-merge can enqueue it.

brandon-coproduct:
Your single failure is the workspace line-coverage floor, and it is **0.27 percentage points** short — but the floor has almost no headroom to begin with, so this is worth context before you go hunting for tests.

```
this branch   line coverage 82.73%   (floor 83)
main today    line coverage 83.09%   (floor 83)
```

Main passes by **0.09 percentage points** — about **140 lines** of a 155,641-line workspace. I confirmed that from #2769, which is 0 commits behind main and adds no code of consequence: it measures exactly `83.09%`. So any PR adding roughly 140 uncovered lines reds this gate, and this one adds 31 files.

## A note on reading the output

The `TOTAL` row is **regions / functions / lines**, and the *line* figure is the last one. The first percentage is region coverage. I misread it myself for two passes and told someone their 83.34% "should pass" when their line coverage was 82.82%. If you check your own number, take the third percentage.

## What this probably means for you

Not that the branch is wrong — that content-addressed pod identity is 31 files of new code and the floor is set at the measured value with no slack. Either the new code needs tests to carry its own weight, or the floor needs headroom restored deliberately. The second is a repo-level decision, not yours to make in this PR.

Recorded as gatehouse `FINDINGS.md` F-27, alongside the clippy cast ratchet, which is at `345 against a ceiling of 345` — the same zero-headroom condition, and four PRs are currently red across the two.

brandon-coproduct:
`a real nucleus pod boots and is enforced (x86_64)` is red on `3e6cf14e`. **It is not this PR's failure, and it is red on `main` too.**

## What is failing

```
Error: run was refused (403) but NOT by the admission gate:
{"error":"kernel denied: no signed admission credential covers it
 (no issuer-signed credential presented for this operation) (operation run_bash on true)",
 "kind":"kernel_denied"}
A refusal for another reason (lattice, scope, IFC) does not prove
the verified-admission gate fired.
```

Everything before that line passed: the pod booted in 1170 ms, the guest proved itself to its proxy, the allowed operation was served, the forbidden one was denied, the leak sweep came back `canary=absent`. The refusal the check is inspecting *did* come from the admission gate — the message says so in words.

## Why it is not this PR's

`crates/nucleus-cli/src/verify.rs:623` on `main` identifies that gate as:

```rust
if !body.contains("DlcAdmissionDenied") {
```

`DlcAdmissionDenied` is a Rust **Debug** variant name. It only ever reached the wire because the proxy formatted `DenyReason` with `{:?}`. #2758 replaced that Debug rendering with a written sentence — which was the right change, and which removed the only token this check was reading. The check has been red on `main` since #2758 merged at 04:45Z.

This PR touches none of it. Its diff is the content-addressed launch path and `nucleus-node` tests; it inherited the failure by merging `main`.

ADR 0007 **I-2** names the shape: *a string compared to a string is not a check.* The gate was keyed on a by-product of a derive, so a presentation change silently became a gate change.

## The fix exists

#2762 adds a machine-readable `deny_code` to `ErrorBody` and points `verify.rs` at `dlc_admission_denied` — a code chosen deliberately rather than scraped out of Debug output. I have just merged current `main` into it (`21ab657a`) and it is armed for auto-merge; it was blocked on a stale base.

**Not ported into this PR**, deliberately. #2762 is 17 files and ~1283 lines of `ErrorBody`/`ApiError`/SDK work; copying it here would widen this PR with someone else's feature to fix a failure this PR did not cause. Once #2762 lands, this check goes green here on the next run against `main`.

I have not re-run the job — a re-run against today's `main` would fail identically, because the cause is in `main`.

---

PR #2719, opened 2026-09-09
ci(fly): the worker image carries Charon and Aeneas

brandon-coproduct:
The Aeneas lanes install charon/aeneas only when `runner.environment == 'github-hosted'`; on the self-hosted fleet they assume the image has them, and it did not — four required contexts failed across #2685, #2689 and #2690. Third instance of the class #2718 pins.

Trimmed against `docker/Dockerfile.runner`: that image installs Charon's nightly with miri and seven cross targets, which is most of what makes it exceed Fly's 8 GB rootfs cap. Only the host target is needed here.

Image 1.8 → 2.5 GB, canaried on one machine before the fleet — no unpack failures.

https://claude.ai/code/session_01JYc8hPky7z1FvG57wAaRMc

### DIFF
--- a/ci/fly-runner/Dockerfile
+++ b/ci/fly-runner/Dockerfile
@@ -113,6 +113,29 @@ RUN set -eux; \
     /tmp/elan-init -y --no-modify-path --default-toolchain "${LEAN_TOOLCHAIN}"; \
     rm -f /tmp/elan-init; lean --version; lake --version
 
+# Charon + Aeneas, the Rust → Lean extraction lanes. Same reason as elan above: the aeneas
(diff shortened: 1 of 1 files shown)


brandon-coproduct:
Holding this: the image it builds is what blew Fly's 8 GB uncompressed rootfs cap. 24 machines logged `Not enough space to unpack image` and wedged in `starting`; the fleet is rolled back to the 1.78 GB Lean image (`sha256:1eae0505…`). #2721 adds the build-time refusal that would have caught this. Charon/Aeneas needs to come from a volume, not the rootfs.

PR #2737, opened 2026-09-09
ci(aeneas): install elan on any runner that lacks it, not only hosted ones

brandon-coproduct:
Both Aeneas workflows gated the elan install on `runner.environment == 'github-hosted'`, so on the self-hosted pool they assumed every machine carries a baked elan.

That assumption broke tonight. The **same workflow**, on the **same commit**, with **no workflow changes on the branch**:

| run | runner | `lake build (extracted core + theorem)` |
|---|---|---|
| main @ `f3f5877` | `nucleus-fly-build-9` | ✅ success |
| PR #2690 | `nucleus-fly-build-4` | ❌ **exit 127** |

Same label (`nucleus-fly-build`), same runner group. A pool is not homogeneous just because its label is — after the fleet was rebuilt following the #2719 rootfs incident, some machines came back without elan.

## The failure was also mis-signposted

The self-hosted branch runs:

```yaml
run: lake exe cache get || true
```

That `|| true` swallows the *first* missing-`lake` error, so the run dies one step later on a bare `Process completed with exit code 127` with no mention of `lake` anywhere. It reads as a broken Lean proof rather than a missing toolchain — which is exactly how it presented on #2690, alongside a wall of unrelated "generated Funs.lean drifted" warnings.

## Change

The step now runs on **every** runner and no-ops when `lake` is already on PATH:

```bash
if command -v lake >/dev/null 2>&1; then
  echo "lake already present: $(command -v lake)"
  exit 0
fi
```

So it costs one `command -v` where elan is baked, and self-heals where it isn't. Same pinned elan v4.2.4 tarball and same SHA-256 check as before — no change to the supply-chain posture (Scorecard Pinned-Dependencies still satisfied).

This is the practice [GitHub's own self-hosted runner guidance](https://docs.github.com/en/actions/reference/runners/self-hosted-runners) and [AWS's runners-at-scale writeup](https://aws.amazon.com/blogs/devops/best-practices-working-with-self-hosted-github-action-runners-at-scale-on-aws/) both recommend: keep install steps in the workflow so it is runner-agnostic, rather than depending on image contents staying in sync.

## Verified

```
actionlint 1.7.12 -shellcheck= -pyflakes= .github/workflows/*.yml   clean
```

Same version and flags `manifest-guards.yml` runs.

## Scope

Applied to both workflows carrying the gate — `aeneas-ifc-scoped.yml` and `aeneas-certchain.yml`. Unblocks the Aeneas checks on #2690 and #2685, which are otherwise at the mercy of which machine they land on.

I left `lake exe cache get || true` alone: with the guard above, `lake` is always present by then, and a cache-fetch failure genuinely shouldn't red the build.

https://claude.ai/code/session_01E4LNSNppsdYdWBh76xYaGW

---
_Generated by [Claude Code](https://claude.ai/code/session_01E4LNSNppsdYdWBh76xYaGW)_


brandon-coproduct:
Closing as subsumed by #2739, which is now queued.

The two changes are substantively identical — same `command -v lake` guard, same removal of `if: runner.environment == 'github-hosted'` — but #2739 covers **13 workflows** where this covers 2. I diffed both hunks against each other: the extra lines here are comments, not behaviour.

One thing here is worth keeping and #2739 does not say: the *why*.

> "runs on every runner and no-ops where lake is baked in, because the self-hosted pool is not uniformly carrying elan any more."

If someone touches those steps again, that sentence is the reason they exist. Worth a comment on the block in a follow-up.

For the record on the wider choice: #2685 proposed replacing these installs with `leanprover/lean-action`, which is the better structure — it removes 13 duplicated blocks, a bespoke first-party cache action, and the hosted/self-hosted branching. It is on hold because that action's `install_elan.sh`, at the SHA it pins, is `curl .../elan/master/elan-init.sh | sh` — the mutable-ref pipe main deliberately removed, with no input to override it. Recorded as gatehouse FINDINGS.md F-23. The separate fix #2685 identified — cache failures silently falling back to cold builds — is being taken without the mechanism.

PR #2739, opened 2026-09-09
ci(lean): install elan on any runner that lacks it, across all Lean workflows

brandon-coproduct:
Follow-up to #2737, which fixed **two** workflows. It turns out to be **twelve**.

Thirteen workflows install elan, and twelve gated it on `runner.environment == 'github-hosted'` — so each assumed every machine in the self-hosted pool carries a baked elan. That stopped being true when the fleet was rebuilt after the #2719 rootfs incident:

| run | runner | `lake build` |
|---|---|---|
| main | `nucleus-fly-build-9` | ✅ |
| #2690 | `nucleus-fly-build-4` | ❌ **exit 127** |

Same commit, same label, same runner group. A pool is not homogeneous just because its label is.

**What showed the first fix was too narrow:** after #2737, #2690 failed the same way in a *third* workflow, `aeneas.yml`. Rather than chase them one incident at a time, this covers all thirteen.

## Why the failure reads as a broken proof

Wherever the self-hosted branch runs `lake exe cache get || true`, that `|| true` swallows the *first* missing-`lake` error. The run then dies a step later on a bare `Process completed with exit code 127` that never mentions `lake`, sitting next to a wall of unrelated "generated Funs.lean drifted" warnings.

## Change

```bash
if command -v lake >/dev/null 2>&1; then
  echo "lake already present: $(command -v lake)"
  exit 0
fi
```

One `command -v` where elan is baked, self-healing where it isn't. Same pinned elan v4.2.4 tarball and SHA-256 check — supply-chain posture unchanged (Scorecard Pinned-Dependencies still satisfied).

`uniform-primitive-lean.yml` already installed unconditionally, so the pattern had precedent here; it gains the guard so a baked image skips the download it was doing every run.

## Scope conditions preserved

A step keeps its `steps.scope.outputs.relevant` clause and loses only the hosted one. Steps whose *only* condition was the hosted clause become unconditional. I verified after the rewrite that no elan step still gates on `github-hosted` and that every scope condition survived — that check is what caught my first pass reading `aeneas-ifc-scoped.yml` as unconditional when it had in fact kept its clause.

## Verified

```
YAML parse: all 73 workflows OK
actionlint 1.7.12 -shellcheck= -pyflakes=: clean
```

Same version and flags `manifest-guards.yml` runs.

## Flagged, not fixed

`ifc-lean.yml` installs elan by piping `elan-init.sh` from raw.githubusercontent into a shell, while the other twelve use the checksum-verified tarball *specifically so no downloaded script is handed to an interpreter*. That inconsistency predates this change and belongs in its own PR.

Also corrects two comments that now said the opposite of the code ("Hosted only: the runner image bakes elan").

Unblocks the Lean/Aeneas checks on #2690 and #2685, which are otherwise at the mercy of which machine they land on.

https://claude.ai/code/session_01E4LNSNppsdYdWBh76xYaGW

---
_Generated by [Claude Code](https://claude.ai/code/session_01E4LNSNppsdYdWBh76xYaGW)_

### DIFF
--- a/.github/workflows/aeneas-certchain.yml
+++ b/.github/workflows/aeneas-certchain.yml
@@ -173,9 +173,14 @@ jobs:
           done
           echo "generated-certchain/ canonicalized from the fresh extraction; theorem builds against it."
 
-      - name: Install elan (Lean toolchain)
-        if: runner.environment == 'github-hosted'
+      - name: Install elan (Lean toolchain) unless the runner already has it
+        shell: bash
         run: |
+          if command -v lake >/dev/null 2>&1; then
+            echo "lake already present: $(command -v lake)"
+            exit 0
+          fi
+          echo "lake not on PATH; installing elan"
           # elan v4.2.4 as its release binary, checksum-verified before it runs; no downloaded
           # script is handed to an interpreter (Scorecard Pinned-Dependencies).
           curl -sSfL https://github.com/leanprover/elan/releases/download/v4.2.4/elan-x86_64-unknown-linux-gnu.tar.gz -o /tmp/elan.tar.gz
--- a/.github/workflows/aeneas-ifc-scoped.yml
+++ b/.github/workflows/aeneas-ifc-scoped.yml
@@ -436,9 +436,15 @@ jobs:
           echo "CANONICAL-FUNS-END"
 
       # --- Build the theorem + capture the REAL axiom set ---
-      - name: Install elan (Lean toolchain)
-        if: steps.scope.outputs.relevant == 'true' && runner.environment == 'github-hosted'
+      - name: Install elan (Lean toolchain) unless the runner already has it
+        if: steps.scope.outputs.relevant == 'true'
+        shell: bash
         run: |
+          if command -v lake >/dev/null 2>&1; then
+            echo "lake already present: $(command -v lake)"
+            exit 0
+          fi
+          echo "lake not on PATH; installing elan"
           # elan v4.2.4 as its release binary, checksum-verified before it runs; no downloaded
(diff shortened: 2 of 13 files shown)


brandon-coproduct:
Rebased onto main. The conflict was **not** mechanical, and taking either side whole would have lost a fix.

`main` had meanwhile hardened `ifc-lean.yml`'s elan install: pinned release tarball at `v4.2.4`, SHA-256 verified, explicitly replacing `curl .../elan/master/elan-init.sh | sh` because `master` is a **mutable ref** — "the bytes CI executed were whatever that branch held at the moment of the run, and the pipe left no point at which they could be checked."

This branch predates that, so its body for that file was the old pipe. A naive `--theirs` would have **reintroduced the supply-chain regression main had just fixed**.

Resolved by keeping both intents: this PR's `command -v lake` guard wrapping **main's** pinned, checksum-verified install. The same reasoning applied to `portcullis-core-proven-lean.yml`, where main had also added a "Reject compiler overrides" step this branch's hunk would have dropped.

Verified after the rebase:

| | |
|---|---|
| workflows with the `command -v lake` guard | **13** |
| workflows installing from the pinned tarball | **13** |
| workflows running `sha256sum -c` on it | **13** |
| real `curl … elan/master … \| sh` pipes remaining | **0** |

Also worth recording: before the rebase this PR's only failures were the three **informational** gatehouse shadow gates. Every required check passed.

Note for whoever reviews #2737 — its change is substantively **identical** to this one (same guard, same removal of `if: runner.environment == 'github-hosted'`), over 2 files where this covers 13. The extra lines in its diff are comments. It looks superseded, though its comment on *why* the pool is no longer uniformly carrying elan is worth keeping.

Issue #2738, opened 2026-09-09
exec escapes the label system for a clean session: make the threat-model boundary explicit and offer a no-exec posture


bcrisp:
Everything below is already documented in the repo. The request is to make the
**threat-model boundary** explicit and to offer a composed posture for the case
the current documentation does not address: a model that is itself adversarial
and can read this source.

## The mechanics, stated accurately

`ifc_ops.rs:508` pins the floor, and its own message draws the line:

```rust
fn bash_exec_floor_gates_the_spawn_not_the_reach() {
    assert_eq!(SinkClass::BashExec.required_integrity(), IntegLevel::Untrusted,
        "BashExec must refuse an adversarial-tainted session from spawning a shell; \
         where a running shell REACHES is confined by the netns backstop \
         (check-egress-probe.sh), not this label");
}
```

So an **adversarially-tainted** session cannot spawn a shell at all. That is the
injection case, and it is handled: the label system refuses the spawn.

A **clean** session can spawn one, and after that the label system has no
visibility into what the shell does. `production-delta.md` says it plainly —
one `(RunBash, BashExec)` authority buys arbitrary syscalls in the guest and
`bash -c 'curl | sh'` never reaches `NetEffect::fetch`. Containment for that
surface is the netns/iptables default-deny and nothing else.

Command *output* is taint-classified by `ingest.rs:189`
(`command_output_is_external` → `portcullis::egress::analyze_egress`). That
function is well built — it handles pipelines, path forms, interpreter
indirection (`python -c` counts as `ScriptingEgress`), and is fail-closed on
input `shell_words` cannot split. It also states its own bound:

> **Bound, stated plainly:** this is a heuristic over command text, and
> obfuscated program names can evade it. It is defence in depth, not the
> containment — that is the pod's default-deny egress.

## Why the threat model matters here

Under **indirect prompt injection**, the session carrying the attacker's
instructions is Adversarial, so `BashExec` refuses the spawn. The design is
sound and the proofs are load-bearing.

Under an **abliterated or intentionally malicious model**, the session is clean
by construction — the operator's own prompt is Trusted/Directive. Nothing has
tainted it. So the model spawns a shell legitimately, and from that point:

* the IFC lattice sees nothing it does;
* the taint classifier is a documented-evadable heuristic *that the model can
  read*, and it only decides whether output is labelled, not whether the command
  runs;
* the actual perimeter is the pod's netns default-deny, whose semantics are
[...]

PR #2745, opened 2026-09-10
docs(threat-model): name the adversary the lattice does not cover, and give it a profile

brandon-coproduct:
Addresses #2738 items **1** and **2**. Items 3 and 4 are named in the doc as open, not closed.

## The distinction

The non-interference guarantee has an adversary in mind, and which one it is changes what the runtime is doing for you. That was already true and already recorded — in a test name and a row of `production-delta.md`. Not where someone deciding whether to trust the runtime will look.

**The confused deputy** — covered. The model works for you, an attacker gets content in front of it, the mediated ingest makes that session `Adversarial`, and the `BashExec` floor refuses the spawn. No shell starts; there is nothing downstream to contain.

**A hostile principal** — abliterated, backdoored, or misaligned, and able to read this source — not covered, and the reason is precise: its session is **clean by construction**. The operator's own prompt is `Trusted`/`Directive`, nothing tainted it, so the lattice permits the spawn. After that the lattice sees none of the shell's syscalls: one `(RunBash, BashExec)` authority buys arbitrary execution in the guest, and `bash -c 'curl | sh'` never reaches `NetEffect::fetch`. The perimeter for that surface is the pod's netns default-deny — a real boundary, and a much narrower claim than non-interference.

## What's here

1. **README**, beside the lethal-trifecta claim. The qualifier already there is about the *channel* (which bytes are observed); this one is about the *adversary*, and the two are independent.
2. **`docs/adversarial-model-posture.md`**, linked from `SUMMARY.md`, working it through.
3. **`profiles/untrusted-model.yaml`** — the composed posture the issue asked for: `run_bash: never`, `spawn_agent: never`, `manage_pods: never`, and **every** egress sink `never`. A model that cannot spawn a shell but can open a PR is not contained. Reads, writes, edits and search remain, so it's a profile somebody would actually adopt.

The reasoning is capability removal rather than flow control, and that's the point rather than a shortcut: a flow-time denial fires only when the session is tainted, and this adversary's session never is. **Arity is the defence.**

## One ask that turned out to be unnecessary

#2738 wanted this posture to default `NUCLEUS_PARANOID_TOOL_IO=1` so no heuristic sits in the path. Under `run_bash: never` **there is no heuristic to default**: `/v1/run` maps to `Operation::RunBash` at the run gate (`run_gate.rs:636`), `RunBash` isn't in scope, so the request is denied before any command runs and `should_observe_command_output` is never consulted for that pod. Out of the path by construction rather than by configuration.

It's still worth setting for the five profiles that *do* permit exec (`codegen`, `local-dev`, `test-runner`, `release`, `safe-pr-fixer`), where the classifier's stated evasion is a white-box target. Making it per-profile is item 3 — not done here, and the doc says so.

## Tests

- `untrusted_model_removes_the_exec_and_egress_surfaces` asserts **capabilities, not verdicts** — deliberately, because against this adversary a verdict-level assertion would be checking a denial that never fires. Carries a non-vacuity control that the profile isn't simply `Never` everywhere.
- `codegen_permits_the_exec_untrusted_model_removes` is the contrast: if `codegen` ever became `Never` the new profile is redundant; if `untrusted-model` ever became `LowRisk` the two are indistinguishable.
- The provider count moves 10 → 11; `build.rs`'s doc comment stops hard-coding the number.

## Local gates

`cargo test -p portcullis` — 1028 passed / 0 failed · `cargo clippy -p portcullis --all-targets --all-features -- -D warnings` (CI's exact invocation) · `cargo fmt --check` · `scripts/check-line-ratchet.sh` · `ci/no-vendor-strings.sh`

**Noted while verifying, not fixed here:** `cargo clippy -p portcullis --all-targets` at *default* features is already red on `main` — four unused-import errors in `tests/adversarial.rs` and `tests/proptest_math_framework.rs`, whose imports are used only inside `cfg(feature = ...)` blocks. CI runs `--all-features`, so it never sees them. Filed separately.


brandon-coproduct:
Pushed a fix for the one red check.

`Code Coverage (llvm-cov)` was failing on a **real test**, not the threshold:

```
profile::tests::test_canonical_profiles_parse
assertion `left == right` failed
  left: 11
 right: 10
```

`assert_eq!(registry.len(), 10)` at `profile.rs:655` — a third hard-coded copy of the number of files in `profiles/`. I updated the other two and missed this one.

**Why I missed it is the part worth stating**, because it isn't carelessness about a grep: this test *does not exist at the default feature set*. `cargo test -p portcullis --lib` compiles and runs 1028 tests; with `--all-features` it runs 1200, and this is among the extra ones. Filtering for it by name at default features reports `0 passed; 0 failed; 1028 filtered out` — which reads exactly like a pass. My local run was green about a test it never compiled.

Rather than change 10 to 11 and leave a fourth copy waiting for the next profile, the count is now derived from the provider:

```rust
assert_eq!(registry.len(), ProfileName::ALL.len());
```

Not vacuous — `canonical()` registers *by name*, so two profiles claiming one name would collapse and make `len()` short. That asserts what the test was actually for, without restating a number that lives in the filesystem. `untrusted-model` joins the explicit name list beside it, which is what really pins membership.

Re-verified with the invocation that matters this time: `cargo test -p portcullis --all-features` — 1200 + 36 + 24 + 6 + 2 + 8 + 6 + 12 passed, 0 failed.

Worth noting alongside #2746: clippy on this crate is red at *default* features and green under `--all-features`, and the tests are the exact opposite. Neither invocation alone covers portcullis.

Issue #2406, opened 2026-09-03
A granted approval does not satisfy the retry it was granted for


brandon-coproduct:
Against a live pod: `/v1/approve` accepts an Ed25519-signed approval (**200**), and the immediately following write for the *same* operation string is still refused `approval_required`.

```
unapproved write refused       403   ok      <- correct, the gate works
approval accepted              200   ok      <- /v1/approve returns 200
approved write succeeds        403   FAILED
  {"error":"approval required: 'WriteFiles perf-163828-3427016852.txt' requires human
   approval","kind":"approval_required","operation":"WriteFiles perf-163828-3427016852.txt"}
```

The operation string is **echoed from the first 403 body**, not reconstructed, so it is byte-identical to what the write path reports. Approvals key on it verbatim (`ApprovalRegistry::approve` / `consume`).

## Hypotheses tested and refuted

1. **Double consumption.** `write_file` consumes in its guard (`|| state.approvals.consume(&op)`) and then calls `request_approval`, whose installed approver is `move |request| approvals.consume(request.operation())` — two consumes per attempt. *Refuted:* granting `count: 2` changes nothing.

2. **Two registries.** `AppState` is constructed with the same `Arc<ApprovalRegistry>` that is closed over by `CallbackApprover` (`main.rs:1484` → `1848`). One registry.

3. **Key transformed on the way in.** `approval_request` is `ApprovalRequest::new(operation)` and `operation()` returns it unchanged, so the approver looks up the same string.

4. **Immediate expiry.** With no `expires_at_unix`, `resolve_approval_expiry` returns `now + MAX_APPROVAL_TTL_SECS`, and `is_expired` is `ts <= now`. Not expiring.

## Why the error message cannot narrow it further

Both the guard failing and the approver failing produce the *same* `ApprovalRequired` with the same operation string, so the response cannot distinguish "the approval was never found" from "it was found and the second consume failed". That ambiguity is most of why this took four refuted hypotheses — the two branches should be distinguishable from the outside.

## Repro

```
nucleus-perf toolcall --spec <pod.yaml> --auth-secret "$SECRET" \
  --approval-key /var/lib/nucleus/state/approval_signing_key.der
```

aarch64 / KVM, node and rootfs built from the same tree.

## Note on what IS working

[...]

PR #2754, opened 2026-09-10
feat(agency): measure the numerator — first MEASURED reading of the frontier (ADR 0005)

brandon-coproduct:
Stacked on #2752 (the #2406 fix), which it depends on.

## What this is

Every number nucleus publishes is a proof count, a denial count, or a latency. All of them are about the constraint; none is about what the constraint is *for*. [ADR 0005](docs/adr/0005-delegatable-agency.md) states the objective as a ratio and says plainly that its numerator has never been measured. This is the instrument, and the first reading.

**First MEASURED reading** — `codegen` profile, Tier 2 microVM, aarch64/KVM:

```
agency: 5/5 tasks completed (100%) under microvm enforcement
cost:   ρ_effect = undefined · ρ_dimension = 1.75 · C(T) = 4 ·
        7 denial(s) inside the grant of 8 total, 5 of them deferrals to a person ·
        risk Medium
valid:  3 containment check(s) held
```

**It started at 3/5.** The instrument found three defects in the runtime on its first run, and the rate moved as each was closed:

| reading | rate | what was in the way |
|---|---|---|
| 1 | 3/5 | the kernel mediated an overwrite as `WriteFiles` while the sandbox enforced `EditFiles`; the command executor keyed approvals on the raw command (`echo hello`) — a **third** vocabulary |
| 2 | 3/5 | both named correctly, but the run guard *spent* the grant the executor's own approver then needed |
| 3 | 4/5 | the edit's discharge authorised `EditFiles` while `Sandbox::write` still spent as `WriteFiles` |
| 4 | 5/5 | the spend follows the operation |

All four were one rule at four depths: **one act, one operation, one name, at every gate.** None was visible to a unit test, because each gate was internally consistent and only disagreed with its neighbour. The denominator was green throughout — which is the argument for measuring the numerator.

## Pieces

- **`portcullis::agency_report`** — the object that holds both halves. Deliberately *not* derivable from a trace: a run in which nothing was refused is not evidence that anything was achieved.
- **`UsageReport::denials_within_grant`** — the friction term. Denials *outside* the grant are excluded on purpose: those are the boundary doing its job on work nobody authorised, and counting them would make a tighter grant score worse. Documented as an upper bound on wrongly-refused work, not a bug count.
- **`nucleus-perf agency`** — five work tasks, three containment checks, against a live pod. Tasks are Rust, not a manifest DSL, because real work has data flow in it (write *these* bytes, read *that* path back, compare) and a DSL that cannot express that would measure a toy.
- **`approval::approval_key`** — the naming rule, now living with `ApprovalRequest`; the sandbox and the command executor forward to it.
- **`Sandbox::write_operation_for`** — exposes the write-vs-edit test the sandbox already makes, so the monitor upstream decides about the operation that will actually be enforced.

## Three properties pinned by tests, because each is a way the number could lie

- **Containment is not work.** Refusal checks are excluded from the numerator. A suite that counted its own refusals as completions would make ℐ *rise* as the runtime got more restrictive — inverting the objective.
- **A breach invalidates rather than subtracts.** One failed containment check voids the report; there is no exchange rate between "did more work" and "the boundary held". An empty containment set is also invalid — a suite that did not look has established nothing.
- **An empty suite has no completion rate.** `None`, not 1.0. 0/0 rendered as 100% would let a harness that ran no tasks pass a ratchet forever.

`Enforcement` is stated by the runner, never inferred: "5 of 5 under a microVM" and "5 of 5 with the gate off" are the same fraction and not the same fact.

## Invariants kept

No default widened, no ceiling raised. Both approval changes narrow or clarify: the naming rule cannot admit anything a correctly-spelled grant would not have, and the peek/spend split reduces the number of grants an operation consumes from two to one. `a_grant_cannot_move_a_denial` still pins that a grant settles a deferral and nothing else.

## What the reading does not establish

Stated on the page itself, not in a footnote: it is not a model benchmark; five tasks is a floor; one profile, one architecture, one run, and no unconstrained control arm; and **ρ_effect — the number that actually matters — is missing**, because this pod ran under a profile and ρ over effects needs a compiled grant to divide by. Reported as null rather than 1.0.

## Validation

`cargo test`: portcullis 1211, nucleus 61, nucleus-tool-proxy 421, all passing (17 new). Clippy `-D warnings` clean on all four crates; `cargo fmt --check` clean; `check-mediation.sh`, `check-line-ratchet.sh --strict`, `ci/no-vendor-strings.sh`, `scripts/formal-numbers.sh` all pass. Live: four `nucleus-perf agency` runs on a real Firecracker pod, the last 5/5 with containment intact.


brandon-coproduct:
## Part C: four more basis vectors (`ececf02`)

Effects are basis vectors of authority space (ADR 0005, decision 5), and the catalog spanned five — `fs`, `shell`, `git`, `web`, `github`. Anything a person might want delegated *outside their own repository* had no name at all: a goal about a cluster compiled to nothing.

Four packs — `aws`, `kubernetes`, `database`, `slack` — each split where a **person** would split, not where the API does. "read pod logs" and "roll out a workload" are different asks; `kubectl` does not make them look different.

### The pack's own conformance test found a hole in the pack

With only exact hosts and a leading `*.`, no regional endpoint was expressible, so every AWS effect had to claim `*.amazonaws.com` — and an effect with a host list and no discriminating HTTP shape vouches for its hosts. So the first draft let a grant of *"list cloud resources"* admit a `POST` to `iam.amazonaws.com`: the one call that rewrites the boundary itself.

`host_matches` now takes a `*` in any label position, matching one or more whole labels; the leading case is bit-identical to before. It is **the egress gate's matcher too**, so the anchors get their own adversarial tests — `*.amazonaws.com` admits neither `evil-amazonaws.com` nor `amazonaws.com.evil.example`, a `*` never matches zero labels, and a `*` is a whole label or it is not a wildcard.

### Compiler rules propose reads only

A goal is evidence about what a person wants to **learn**, and much weaker evidence about what they will let be **changed**. *"the deploy is broken"* plausibly means read the logs; it does not mean roll something out, and a rule that inferred so would widen authority from a guess.

Six goals naming a deploy, a migration, a message and an IAM change are checked to propose **none** of the twelve mutating effects — with a companion test that those same goals *do* fire their rules, so the absence is a decision rather than a failure to match. Mutations stay reachable by being asked for (`--effects`) or via the escalation proposal the read-only grant's denial produces. Both put it in front of a person.

### It works end to end

```
$ nucleus run --goal "check the cloudwatch logs for the lambda" --dry-run --ceiling research-web
Can:     list cloud resources · read CloudWatch logs · read and search workspace files
Limits:  $1.50 · 44m · *.s3.amazonaws.com, ec2.*.amazonaws.com, ecs.*.amazonaws.com,
         lambda.*.amazonaws.com, logs.*.amazonaws.com, rds.*.amazonaws.com,
         s3.amazonaws.com, sts.amazonaws.com only
```

Under a ceiling that does not admit egress, the same goal renders those effects in `Cannot` **with the reason** (`outside ceiling local-dev: web_fetch is never`) rather than failing to recognise the goal. A denial that names what it would take is the raw material of an escalation proposal; silence is not.

### No measured number moved, and the page says so

The packs widen what *can* be delegated; the suite measures what one pod *did*, which was filesystem and shell work under a profile. Seeing them in a number needs a `--goal` grant (so ρ_effect is defined at all) and tasks that touch those surfaces. Staging a delta would have been easy and dishonest.

Also fixes a feature gate this branch broke one commit earlier — the `agency_report` re-export took the `cfg` belonging to `grant_usage`, so the crate failed to build without `spec`.

**Validation:** portcullis 1223 tests (11 new), nucleus-task-compiler 13 (5 new), nucleus-tool-proxy 421, nucleus 61 — all passing; clippy `-D warnings` and `fmt --check` clean; `check-task-compiler-offline.sh` and `no-vendor-strings.sh` pass.

brandon-coproduct:
## ρ_effect exists (`d20469f`)

The number the effect catalog is supposed to move was `null`, for a structural reason rather than an oversight: ρ over effects divides granted effects by exercised ones, and a **profile has no semantic effects to divide by**. Only a compiled grant does.

`nucleus-perf agency --local` compiles a goal, seals it, and spawns a tool-proxy under the sealed certificate — same flags, Tier-3 orchestrator token and announce-file handshake `nucleus run --local` uses, so it measures the shipped local path rather than a parallel copy. The grant's effects are sealed into the certificate the proxy verifies, so the run is **bounded by** the effects it is **measured against**.

**MEASURED, Tier 1 local, macOS arm64:**

```
agency: 5/5 tasks completed (100%) under local enforcement
cost:   ρ_effect = 1.25 · ρ_dimension = 1.75 · C(T) = 1 ·
        3 denial(s) inside the grant of 4 total, 0 of them deferrals · risk Medium
valid:  3 containment check(s) held
```

Two things there matter more than the completion rate:

- **ρ_effect = 1.25** — the grant holds five effects, the run exercised four. One went unused, and `nucleus observe --narrow` would propose dropping it. That is the ADR 0004 loop closing on a measured number for the first time.
- **C(T) = 1, against 4 under the profile.** The same five tasks cost four in-run approvals under `codegen`, which defers every write, edit and shell call. Under the grant: one confirmation before anything ran, zero during. That confirmation is **counted**, not dropped — the grant is one decision instead of four, made up front rather than four times while a person waits.

| | Tier 2, profile | Tier 1, compiled grant |
|---|---|---|
| work completed | 5/5 | 5/5 |
| ρ_effect | *undefined* | **1.25** |
| C(T) | 4 | **1** |
| enforcement | microVM | in-process |

Not a controlled comparison — different tiers, hosts, architectures — and the page says so. What the pair shows is the shape ADR 0004 predicted.

### Three things surfaced on the way

1. **The local proxy refuses unsigned requests**, and it showed up in the worst-looking way: every task failed *and* a containment check "failed", so the reading refused to be quoted. The harness couldn't tell an auth refusal from a policy one — which is what `is_valid` exists for.
2. **`nucleus_client::sign_http_headers` does not verify against the tool-proxy.** It signs `{ts}.{actor}.{nonce}.{body}`; `auth::verify_http` reconstructs `{ts}.{actor}.{body}` — its own doc comment says so. The node accepts the nonce-bearing form, so the two verifiers disagree and the shared helper matches only one. **Flagged, not fixed:** `nucleus-mcp`, the shipped bridge from an agent to the proxy, signs with that same helper (`nucleus-mcp/src/main.rs:360`). Not asserted as broken — this harness hasn't run that path end to end — but it's the #2406 shape one layer out, and nothing compares the two producers.
3. **The discharge gate needs a session task token.** Without one `verified_scope` is `None` and `InScopeWithTask` denies everything fail-closed — correct, and it read as 1/5 with "no verified scope present" on every task.

### A containment check was wrong, and the mechanism caught the *suite*

`refuses-unapproved-write` asserted an unapproved write is *deferred* — true under `codegen`, false under a compiled grant for the same work, which authorises writes outright and says so in the grant a person reads (`no approval prompts expected`). The write landed, the check reported a breach, the report refused to be quoted. **The check was the thing that was wrong**, encoding a property of one profile rather than an invariant. It is now `refuses-write-to-a-blocked-path`, which holds under every profile and every grant.

Recorded in the frontier page rather than quietly swapped: containment checks decide whether a whole reading may be quoted, so one that depends on which grant is in force can void a good measurement — and this is the validity property working in the direction nobody designs for.

Both readings now live under `benchmarks/agency/`.

PR #2756, opened 2026-09-10
feat(federation): a token exchange that attenuates

brandon-coproduct:
## What this is

`scope` was echoed from the request verbatim — `token.rs:343` was `scope: req.scope.clone()`. A workload asked for a scope and the OP minted it.

The federation rule bounded **which** audience a subject may reach and **for how long**. Nothing bounded **what the token may do when it gets there** — so the delegation ceiling that the kernel, the certificate and the effect gate all enforce inside the boundary stopped at the one place a federated credential most needs to carry it.

Nothing downstream enforces on our `scope` today, which made it inert rather than dangerous. But it is precisely the claim a relying party keys on — [Snowflake](https://docs.snowflake.com/en/user-guide/workload-identity-federation) (GA July 2026) and [Vault](https://www.hashicorp.com/en/blog/implementing-workload-identity-with-hashicorp-vault-and-spiffe) both federate against a SPIFFE-ID subject and a JWKS-backed issuer today — and the moment one does, an unbounded scope is an unbounded credential.

## `FederationRule::max_scope`

The ceiling goes where the other two ceilings already live, with three states spelled the way TOML spells them:

| `max_scope` | meaning |
|---|---|
| absent | the rule bounds no scope. A request that **asks** for one is refused; a request that asks for none is unaffected. |
| `[]` | constrained to nothing — distinct from absent, the way an empty effect surface is distinct from an unmarked one. |
| `["a", "b"]` | the requested scope must be a **subset**. |

**Refused rather than trimmed.** A caller asking for `read write` under a rule admitting `read` gets an error, not a token that quietly does half of what they asked. A credential that silently means less than its holder believes is its own class of incident — the holder proceeds, the call fails downstream, and nothing points at the scope. And the token grants what was *asked for* rather than the rule's whole ceiling, so asking for less gets less.

## Why absent isn't fail-closed, and why that's still fail-closed

Absent is the migration-safe default, and the distinction is the point: this closes the **hazard** (an unbounded scope being minted) while costing nothing to any caller that never wanted a scope — which is every caller today, since scope is optional in RFC 8693 and nothing enforces on it. No deployment breaks; the hazard closes. Operators tighten by setting `max_scope`, and the runbook says to.

## The response says nothing

The caller sees a bare `invalid_target`. Which scopes a rule admits is operator information, and answering it would make the token endpoint a policy oracle a caller could enumerate — the OP already makes that call for federation denials, and a scope denial is the same class. Refused scopes and the ceiling go to the log at `WARN` with the rule id.

Two of the new tests pin **both halves**: the refusal happens, *and* the response does not enumerate the ceiling. (I had this wrong first — my initial test asserted the body named the refused scope. The code was right and the test was wrong.)

## The conformance doc was under-reporting by four rows

`docs/wimse-aims-conformance-gap.md` still listed `typ`, `client_id`, `scope` and `act` as violations or gaps. All four are closed, and have been for months — `at+jwt` is hard-coded and CI-gated, `act` is set on every exchange per RFC 8693 §4.1. Annotated with what I verified against the code today rather than rewritten, so the original analysis stays readable next to what happened to it.

`cnf` remains deliberately deferred — bearer-only in v1, pending WIMSE WPT settling (those drafts expire between October 2026 and January 2027).

## The gap this does *not* close

Stated in the doc rather than left implied: **a federated token carries the workload's identity, not its authority.** SPIFFE and WIMSE answer *who this workload is*. `max_scope` is the first thing carrying any of *what it may do*, and it is an **operator's** ceiling rather than the pod certificate's own effect dimension.

Deriving the scope from the certificate's effects — so the federated credential cannot outrun the grant a person approved — is the next step and is deliberately not in this change.

## Validation

`cargo test -p nucleus-oidc-provider --all-features`: 116 passing, 7 new. Clippy `-D warnings` clean, `cargo fmt --check` clean, `ci/alg-pin-check.sh`, `ci/no-vendor-strings.sh` and the line ratchet all pass.


brandon-coproduct:
## The principal's ceiling, not just the operator's (`bf1416a`)

`max_scope` above is the **operator's** ceiling: what a rule is willing to issue. It says nothing about whether the workload asking was ever delegated the authority behind it. An operator happy for pods to reach a log service with `logs:write` handed that scope to any pod matching the subject prefix — including one whose grant never included the effect that backs it.

`scope_requires` adds the **principal's**:

```toml
[rule.scope_requires]
"logs:read"  = ["aws/read-logs"]
"logs:write" = ["aws/write-object"]
```

The workload presents its pod certificate as the RFC 8693 `actor_token`; the scope is issued only if that certificate grants those effects. **Both ceilings apply.**

`actor_token` is the right slot rather than a convenient one — RFC 8693 defines it as *the party acting on the subject's behalf*, which is exactly what a pod certificate records. It isn't a JWT and doesn't need to be; the spec lets a token type be any URI.

The map is a translation and has to be: an RP's scopes are its own, nucleus's effects are the units a person granted, and neither side gets to rename the other. The rule is the one place that knows both — and therefore the one place an operator can be asked to state the correspondence deliberately rather than have it guessed.

### The pinned root is the whole security of this

`AttenuationToken::verify` walks the chain against the root key **the token itself carries**. That proves the chain is internally consistent and nothing else — anyone can generate a root and mint themselves a certificate granting `aws/mutate-iam`.

So `NUCLEUS_OIDC_CERT_ROOT_PUBKEY` is compared against the presented root **in constant time, before the chain is walked at all**, and an OP with no pinned root refuses every certificate rather than accepting one on its own say-so. A misconfigured OP issues nothing rather than something it cannot justify.

`a_certificate_rooted_elsewhere_is_refused` covers it, and I mutation-tested it: with the pin disabled the test fails, so it isn't passing for some other reason.

### Six tests, both directions

A backed scope is issued; an unbacked one is refused; a rule demanding backing refuses when no certificate is shown; a foreign-rooted certificate is refused; an OP without a pinned root refuses; and a scope with **no** `scope_requires` entry is unaffected — so the feature is opt-in per scope and adding it breaks nothing.

The refusal stays a bare `invalid_target`, and the tests assert the response does **not** name the effect that would have satisfied it. Which effects back which scope is operator information; answering it would turn the token endpoint into a map of the grant, enumerable one request at a time.

### Dependency note

portcullis enters the OP's tree with `default-features = false, features = ["serde", "crypto"]` — certificate verification and the effect dimension, without Cedar or the policy loader. `effect_surface` is built for exactly this consumer; its module docs already note it stays independent of the catalog so an outside verifier can read what a certificate grants.

### What this buys

A federated credential bounded by **the authority a person actually delegated**, not merely by the identity of the workload asking. SPIFFE says who the workload is; the certificate says what its principal allowed; the token that leaves is bounded by both.

**Still open**, and now recorded in the conformance doc rather than left implied: the certificate rides as a separate `actor_token` rather than inside the issued token, so an RP that wants to re-check the attenuation itself must be handed the certificate too. Emitting the granted effects as a namespaced claim would close that.

**Validation:** 122 passing (6 new); clippy `-D warnings` and `fmt --check` clean; `alg-pin-check.sh` and `no-vendor-strings.sh` pass.

PR #2759, opened 2026-09-10
xtask: measure the ejection rate, because it is what picks a batch size

brandon-coproduct:
Batching multiplies the cost of a red — one failing entry takes every PR batched with it — so the published guidance keys entirely on the failure rate: **under 2% supports batches of 5–10, over 5% wants 2–3.** A queue that doesn't know its own rate can't choose one; it can only guess and find out.

`cargo xtask ci-ejections` reports, per merge-group entry: **merged, ejected, or in flight**, and for an ejection the first *required* check that failed.

## Two things it enforces that measuring by hand got wrong — both of which I did

**In flight is not ejected.** An entry whose PR hasn't merged looks identical to the one sitting at position 1 building right now. Counting the second as an ejection moved my hand figure from **8% to 15%** — across the 5% threshold that decides the batch size. So the live queue is consulted for what it still holds, and only the remainder can be called ejected. If the queue **cannot be read the tool refuses** rather than reclassifying every in-flight entry as an ejection.

**A non-required failure cannot eject anything.** By hand I attributed `#2737`'s ejection to `Analyze (rust)`. That string isn't in `ci/required-checks.txt` — the list has `Analyze` — so it was never required and cannot have ejected anything. Causes are attributed only to contexts in that file, which leaves `#2737` ejected with **no required failure at all**.

That residue is itself a finding: an ejection with no red required check is the signature of a clock rather than a verdict. Two candidates, and I can't distinguish them from what's measured — GitHub's 360-minute entry timeout, or the ENOSPC death that `PoolSpec::requires_volume` documents (*"four required checks failed that way and their merge-queue entries were ejected, which looked from the outside like entries leaving the queue while clean"*).

## It paginates, because the first version couldn't measure anything

A merge group fans out to ~31 workflow runs, so one API page of 100 is barely **three** entries. The first version reported `0/3` and had to shout its own sample size. It still prints `n` and how far one event moves the rate — at n=25 that's 4 points, and the measurement sits just above the threshold it's being used to test.

## Measured now

```
merged 23, ejected 2, in flight 1 (not counted)
ejection rate: 2/25 = 8% of DECIDED entries
batch-size guidance at this rate — over 5%: batches of 2-3
CAUTION: n=25 decided entries is thin. One event moves this by 4 points.
```

**Don't pick a batch size off this yet.** One of the two ejections predates `#2742` (which took CodeQL off the merge-queue path), and the other has no required cause. The tool now exists to re-measure once a clean post-`#2742` window accumulates.

Consistent with the Rust-not-shell mandate: this replaces a shell measurement I ran by hand, and the `Merged | Ejected | InFlight` enum is precisely the distinction shell got wrong.

https://claude.ai/code/session_01MJ65muwqWqPidSYTnpknDi


brandon-coproduct:
The clippy ceiling failure is **three casts**, all in the same expression:

```rust
let rate = 100.0 * ejected as f64 / decided as f64;
                   100.0 / decided as f64
```

CI measures this branch at **346 against a ceiling of 345**.

## Why this reds

main is at **exactly 345 with zero headroom** — confirmed from #2769, which is 0 commits behind main, adds no casts, and measures `345 (ceiling 345)`, passing. The ratchet is shrink-only with `target = 0`, so any new cast in the tracked set reds it. That is the design.

The ceiling was 462 against 458 earlier today; #2771 ground it to 345 with clippy's machine-applicable `cast_lossless` fixes — 117 sites of real debt removed, and no slack left behind.

## The tracked lints care about these specifically

`cast_precision_loss` on `u64 as f64` is the one firing: above 2^53 the conversion is lossy. For a ratio of ejection counts that will never be near 2^53, the honest options are:

- `f64::from(u32::try_from(ejected)?)` if the counts are genuinely small, or
- a scoped `#[allow(clippy::cast_precision_loss)]` with a comment saying the counts are bounded by the queue size — `.clippy-ratchet.toml` names that route explicitly.

Your `Rustfmt` failure is separate and I have not looked at it. The ratchet-falsifier failure is a consequence of this same count: it re-runs the real check as its "green when restored" step.

PR #2762, opened 2026-09-10
feat(denials): a refusal carries the proposal that answers it

brandon-coproduct:
Stacked on #2758. Part B of the DX plan.

## The gap

ADR 0004 milestone 4 built `EscalationProposal` — for a denial it answers the four questions a bare sentence doesn't: what was attempted, why it was stopped, the least authority that would have allowed it, what new risk granting that adds, and the one command that grants exactly the minimum.

It has never crossed a wire. The only constructor, `denials_in_trace`, reads a trace **file** — so the analysis ran at the CLI, after the run had already ended. The agent that was denied got a sentence and gave up; the person got the affordance too late to use it.

## What changed

- `ApiError::Refused { inner, proposal }` — a refusal wrapped with its explanation. `classify()` is split out of `IntoResponse` so the wrapped error's status, kind, operation and message are the inner error's, **unchanged**. A caller branching on `kind` cannot see different behaviour because an explanation became available.
- `proposal` on `ErrorBody`, `skip_serializing_if = "Option::is_none"`.
- `mediation::Denied { error, reason }` — `decide_and_record` hands back the kernel's own `DenyReason` beside the mapped `ApiError`. The mapping is lossy on purpose (several reasons → one stable wire `kind`); a proposal built by guessing backwards from the mapped error would name the wrong minimum authority.
- `proposal::ProposalContext` — grant + ceiling + catalog + cost. Built only from `--pod-grant`, a `SealedTaskGrant` verified against the **pinned** root.
- SDK: `Error::AccessDenied` gains `proposal: Option<Box<EscalationProposal>>`.

## The invariant

A proposal can only ever **add** an explanation. Three tests pin it, one on the actual response bytes; a fourth pins non-vacuity (a refusal built without a grant stays bare).

Absent when there is no grant to propose against — which is every profile run. The key is omitted rather than emitted null, so no client has to distinguish absent from empty.

On the SDK side the same asymmetry: a proposal it cannot decode (newer schema, truncated body, renamed field) leaves the classification exactly where it was. An explanation is a courtesy; the refusal is the contract.

## Verification

Both directions mutation-checked:

| mutation | result |
|---|---|
| `IntoResponse` stops writing the field | `the_refusal_body_carries_the_proposal` FAILED |
| SDK parse always yields `None` | `a_denial_carries_its_proposal_through` FAILED |

`cargo test -p nucleus-tool-proxy -p nucleus-sdk -p portcullis -p nucleus-mcp -p nucleus-mcp-guard --all-features` green (423 + 1207 + …, 0 failures). `cargo check --workspace --all-targets --all-features`, `clippy --all-targets`, `fmt --check`, `check-mediation.sh`, `check-line-ratchet.sh --strict`, `no-vendor-strings.sh` all clean.

## The one judgement call

This widens what an *agent* is told. Justified by its trust position: it is inside the boundary and holds its own certificate, so it can enumerate its grant anyway — and `effect_gate.rs` already tells it `would be admitted by github/open-pr`.

It stays deliberately asymmetric with the token endpoint (#2756), whose caller is a remote workload whose authority is still being decided and which must not enumerate its ceiling. Part F records both, and why, in one place so they don't get "harmonised" later by someone reading only one of them.

## Not in this PR

`deny_reply` in mcp-guard and the `nucleus-mcp` path still stringify. Same shape, separate surface, deferred.


brandon-coproduct:
Sharper diagnosis than my earlier note, and it changes what I think the fix is.

The failing leg is `check_admission_gate` — the **uncredentialed** run, which must be refused with `DlcAdmissionDenied` in the body. What the pod actually returns:

```json
{"error":"kernel denied: no signed admission credential covers it
  (no issuer-signed credential presented for this operation) (operation run_bash on true)",
 "kind":"kernel_denied"}
```

**The reason is right and the `kind` is wrong.** That message is precisely an admission concern — "no issuer-signed credential presented" — reported under a generic kernel kind.

## There is precedent in the same file

`verify.rs:551-556`, on the *other* leg:

> "a booted pod used to report this refusal as `insufficient_capability` with a fabricated `Never`, for a policy that said `Always`. That was the proxy flattening every kernel reason into a capability claim, **since fixed** — the denial now arrives as `kernel_denied` naming the delegation ceiling it exceeded."

This is the same shape one layer along: a specific denial flattened into a generic kind. Last time the fix was in the **producer**, not the check.

## Why widening the check would be wrong

`check_forbidden_operation` deliberately accepts `kernel_denied`, and its comment says why that is safe there. The admission leg cannot do the same: accepting `kernel_denied` would let *any* kernel refusal pass as proof the admission gate fired, which is the one thing this check exists to establish. The file already draws that line explicitly — `approval_required` is excluded for the same reason.

## So

The change wanted is that an admission refusal reports as `DlcAdmissionDenied`, not that the assertion loosens. That is in this PR's own territory — it is the PR making a refusal carry its proposal — so the author is better placed than I am to say whether the kind should be set at the proposal site or upstream.

I also merged `origin/main` into this branch locally to rule out staleness: it merges clean and the failure is unchanged, so being 15 commits behind is not the cause. I have not pushed that merge, since it would re-run ~85 checks for no change in outcome.

PR #2763, opened 2026-09-10
fix(risk): say what the combination permits, and stop claiming the floor is safe

brandon-coproduct:
Part C of the DX plan. Independent of #2758 / #2762.

## The gap

Every disclosure level said the same two things about risk — a grade and a count:

```
Risk:    2 of 3 exposure legs (private data + untrusted content); exfiltration absent → no approval prompts expected
```

Two of three **what**? The sentence a person actually needs — *this combination lets repository data leave the sandbox* — existed at no level, in no renderer.

The material was all there and thrown away. `state_risk` computes the three legs as named booleans and collapses them to a count on the next line. `UninhabitableState` has carried a `name` field for exactly this purpose since it was introduced, and no renderer has ever read it.

## What changed

`exposure_mechanism::mechanism` answers "what does *this* combination permit" for all eight subsets, with no catch-all — a fourth leg fails to compile rather than inheriting the nearest sentence.

Before / after, on real lattices:

```
restrictive  1 of 3 exposure legs (private data); untrusted content and exfiltration absent
             → the agent can read private data, but nothing steers what it reads and nothing carries it out; no approval prompts expected
             trace: risk Low → Low

permissive   all 3 exposure legs present (private data + untrusted content + exfiltration)
             → untrusted content can direct the agent to read private data and send it out of the sandbox;
               the kernel asks for approval before run_bash, git_push, create_pr
             trace: risk Low → Uninhabitable
```

The eight subsets are spelled out rather than composed from per-leg fragments, because the content is in the **joins**. "Can read private data" and "can reach the network" are two facts; "whatever it reads can leave" is a third that exists only when both hold. A generated sentence would have said the first two and lost the one worth reading. There are three two-leg combinations permitting three different things — one *is* the exfiltration path, two are not. That is exactly the distinction the count erased, and it is the property the tests pin.

## Two claims that were made without looking

**`before: StateRisk::Safe` was a literal — and wrong for every caller in the tree.** Both callers weaken from `PermissionLattice::restrictive()`, which holds `read_files`, `glob_search` and `grep_search` at `Always`. That is one exposure leg: `StateRisk::Low`. So every policy trace ever printed opened `risk Safe → X` about a run that started from Low. "Restrictive" describes what the floor withholds, not the floor being harmless.

The plan predicted this would be *unchecked*; measuring it showed it was *false*. `summarise_risk` now takes both lattices plus the cost config and derives the gap and both grades from the same pair, so a trace can no longer describe a weakening from one floor while grading a different one.

**`no approval prompts expected` was inferred from the leg count.** Any grant with fewer than three legs was told flatly that nothing would prompt — false for any grant whose obligations came from another route: an effect pack that gates an operation, a ceiling profile that does, a narrowed grant that inherited one. Such a grant promised no prompts and then prompted. It now reads `approval_gated`.

## Scope

`StateRisk` is untouched. The kernel gates on it and the Lean development reasons about it; this is rendering, deliberately kept out of its way. Both the grade's return type and `is_uninhabitable` are unchanged.

`cargo test -p portcullis -p nucleus-task-compiler --all-features` green (1209 + …, 0 failures). Workspace check, clippy, fmt, `check-mediation.sh`, `check-line-ratchet.sh --strict`, `no-vendor-strings.sh` all clean.

### DIFF
--- a/crates/nucleus-task-compiler/src/compile.rs
+++ b/crates/nucleus-task-compiler/src/compile.rs
@@ -189,10 +189,11 @@ pub fn compile(input: CompileInput<'_>) -> Result<TaskGrant, CompileError> {
         .delegate_to(&lattice, "task grant")
         .map_err(|e| CompileError::NotWithinCeiling(e.to_string()))?;
 
-    let gap = input
-        .cost_config
-        .compute_gap(&PermissionLattice::restrictive(), &lattice);
-    let risk = summarise_risk(&lattice, gap);
+    let risk = summarise_risk(
+        &PermissionLattice::restrictive(),
+        &lattice,
+        input.cost_config,
+    );
 
     let created_at = Utc::now();
     let not_after = lattice.time.valid_until;
--- a/crates/portcullis/src/exposure_mechanism.rs
+++ b/crates/portcullis/src/exposure_mechanism.rs
@@ -1,0 +1,281 @@
+//! What a combination of exposure legs *permits*.
+//!
+//! Every disclosure level said the same two things about risk: a grade, and a
+//! count of legs.
+//!
+//! ```text
+//! Risk:    2 of 3 exposure legs (private data + untrusted content); exfiltration absent → no approval prompts expected
+//! ```
+//!
+//! Two of three what? The sentence a person actually needs — *this combination
+//! permits repository data to leave the sandbox* — existed at no level, in no
+//! renderer. Yet the material was all there and thrown away: [`state_risk`]
+//! computes the three legs as named booleans and collapses them to a count on
+//! the next line, and [`UninhabitableState`] carries a `name` for exactly this
+//! purpose that no renderer has ever read.
+//!
+//! This module keeps the identities. [`mechanism`] answers "what does *this*
+//! combination let the agent do", for all eight subsets, and [`risk_line`]
+//! renders the whole line from a [`RiskSummary`].
+//!
+//! # Why the legs are separate from the grade
+//!
+//! [`StateRisk`] is ordered and load-bearing: the kernel gates on it and the
+//! Lean development reasons about it. Nothing here changes what it computes.
+//! The mechanism is rendering — an English gloss on a set — and is kept out of
+//! the grade's way on purpose.
+//!
+//! [`state_risk`]: crate::IncompatibilityConstraint::state_risk
+//! [`UninhabitableState`]: crate::uninhabitable_state::UninhabitableState
+
+use crate::task_grant::RiskSummary;
+use crate::ExposureLabel;
+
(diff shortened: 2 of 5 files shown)


brandon-coproduct:
Seven red checks, **one cause**, now fixed on this branch.

Every failing job — Clippy, Fuzz, `cargo hack`, Kani toolchain floor, Declassify sink scope, Declassification, Gates-must-fail — was reporting the same eight compile errors:

```
error[E0432]: unresolved import `crate::effect_catalog`
error[E0433]: cannot find `exposure_mechanism` in `crate`
```

## Cause

`lib.rs` on main has `#[cfg(feature = "spec")]` immediately above `pub mod task_grant;`. This branch dropped it while adding the new `exposure_mechanism` block just below it. `spec` is **not** a default feature (`default = ["serde", "crypto", "cedar"]`), so `task_grant` began compiling by default while the modules it imports stayed behind `spec`. rustc said it exactly: *"found an item that was configured out."*

Exactly one module changed — I diffed the gates on all five:

| module | main | this branch |
|---|---|---|
| `task_grant` | `cfg(feature = "spec")` | **ungated** ← |
| `grant_usage` | `cfg(all(spec, not(kani)))` | same |
| `escalation_proposal` | `cfg(all(spec, not(kani)))` | same |
| `sealed_grant` | `cfg(all(spec, not(kani)))` | same |
| `exposure_mechanism` | *(new)* | `cfg(spec)` |

## Verified, not inferred

The check names said "Clippy" and "Code Coverage", which is not where the answer was:

| | |
|---|---|
| without the gate | `cargo check -p portcullis --all-targets` → **exit 101, 8 errors** |
| with the gate | same command → **exit 0** |
| with the gate | `--all-features` → **exit 0** |

The failing invocation uses **default** features. `ci.yml` lints portcullis a second time as `cargo clippy -p portcullis --all-targets` with no `--all-features` — that is what exposes this class, and why #2749 added a gate for it. **That gate did its job here.**

One-line fix pushed, with the reason recorded above the gate so the next person moving a block below it sees why it is there.

brandon-coproduct:
Correction to my previous comment: when I posted it the fix was **not** actually on the branch. My push was rejected as a non-fast-forward — the branch had moved when I ran `update-branch` earlier — and my own script printed "pushed" regardless of the exit code. The diagnosis above was right; the claim that it had landed was not.

It is on the branch now, and I verified it by reading the remote rather than trusting the push:

```
d0baa3e28..fa017c84f  tmp-2763 -> feat/risk-says-what-it-permits
gate present on remote: yes
cargo check -p portcullis --all-targets  →  exit 0, 0 errors
```

Apologies for the noise.

PR #2765, opened 2026-09-10
feat(agency): measure recovery friction — the denominator term nobody was measuring

brandon-coproduct:
Stacked on #2754. Part E of the DX plan.

## The gap

`D`'s denominator is human decisions + configuration + security knowledge + recovery friction. `C(T)` was the only one measured — and it only counts decisions on the **happy path**, the run where the grant was right the first time.

It says nothing about the run where it wasn't. That is where delegation actually fails: an agent is refused something it needed, and either the system tells it what would have worked, or a person goes and reads a profile. And the failure is invisible in a completion rate — the task simply does not complete, with nothing to distinguish "one decision away" from "ten".

## What it measures

`--recovery-goal` runs a deliberately under-granted lane: a read-shaped goal, then a write, so the refusal is the boundary doing its job. From there — refusal read off the wire, `propose` names the least authority that would have allowed it, the grant recompiled with exactly that effect added, the same work attempted again.

First reading, Tier 1 local, ceiling `codegen`:

```
recover: write-after-refusal — refused by kernel_denied,
         proposal named fs/edit-workspace, 1 decision(s) over 0.3s, recovered
```

The target is one decision, and one decision is what it takes.

## What is actually being asserted

Not `decisions == 1`. A proposal that named nothing and a harness that already knew the answer would also score 1. The claim is that the fix came from the **system's** proposal and that it was *sufficient*: refused before, granted exactly the named effect, completed after. The lane is not allowed to know which effect fixes it — the name comes from `propose`, from the catalog and the ceiling.

Both halves were perturbed, and both hold:

| perturbation | result |
|---|---|
| recovery goal wide enough that the write already succeeds | the lane **errors** rather than reporting 0 friction |
| proposal forced to name `fs/read-workspace` instead | `STILL REFUSED after granting the proposed minimum` |

The second row is the one worth keeping. A proposal that names a minimum which does not work is *worse* than proposing nothing: it spends the person's one decision and leaves them exactly where they started. The report has to be able to say so.

## Placement

Recovery is friction, not work. It sits beside `clicks`, never in the numerator, and `recovery` is **absent** rather than zero when the lane did not run — "not measured" and "no friction" are opposite findings. Run with and without the lane, the completion rate, ρ and every denial count come out identical (measured, not assumed).

## Also here

`main.rs` in the tool proxy had crossed its line ratchet (4976 against 4975) on the base branch. Cherry-picked the fix from #2752: `ErrorBody` / `ApiError` / `IntoResponse` move verbatim into `api_error.rs`. No behaviour change; main.rs drops to 4774.

I had been reading that gate's exit code through a pipe, which masked it. Corrected, and re-checked across the whole stack.


brandon-coproduct:
The clippy ceiling failure is **one cast**, and here it is:

```rust
.map(|d| d.as_secs() as i64)
```

CI measures this branch at **346 against a ceiling of 345**.

## Why one cast is enough to red it

main is currently at **exactly 345 with zero headroom** — I confirmed this from #2769, which is 0 commits behind main, adds no casts, and measures `345 (ceiling 345)`, passing. The ratchet is shrink-only with `target = 0`, so at zero headroom *any* new cast in the tracked set reds the gate. That is the design, not a malfunction.

Context worth having: the ceiling was 462 against a count of 458 earlier today. #2771 ground it to 345 by applying clippy's machine-applicable `cast_lossless` fixes. That removed 117 sites of real debt — and left no slack, so this branch is the first to feel it.

## Options

- `d.as_secs()` returns `u64`. `i64::try_from(d.as_secs())` with an explicit failure path is the change the lint is asking for, and here the value is a duration in seconds so the fallible path is nearly dead.
- If the truncation is genuinely fine, a scoped `#[allow(clippy::cast_possible_truncation)]` **with a comment saying why** is the documented escape — `.clippy-ratchet.toml` names that route explicitly.
- Raising the ceiling by one is the third option, but it moves a shrink-only counter the wrong way and wants its own justification.

Your other three failures — the ratchet falsifier, llvm-cov, and Aeneas — I have not diagnosed; the falsifier at least is a consequence of this same count, since it re-runs the real check as its "green when restored" step.

Separately: this branch was **16 commits behind main** and failing against the *old* 462 ceiling. `gh pr update-branch` reported "already up-to-date" while that was true, so I merged main in by hand. It is current now.

brandon-coproduct:
Your last remaining failure is the workspace line-coverage floor, and here is the number rather than the percentage:

```
this branch  129,542 / 156,548 covered = 82.75%   (floor 83.00%)
             needs 392 more covered lines
main         129,327 / 155,641 covered = 83.09%
             margin above the floor:    145 lines
```

So the branch adds **907 lines** to the workspace and covers about **215** of them. The shortfall is not that the new code is untested so much as that main was already sitting **145 lines** above a floor set at its own measured value — any PR adding ~900 lines at below-83% coverage lands under it.

## Reading the output, if you check yourself

The `TOTAL` row is **regions / functions / lines** and the *line* figure is the **third** percentage. The first is region coverage, which reads ~83.3% here and looks like it passes. I misread exactly that for two passes and told you this check "should pass" when it never did — apologies for the detour.

## What this is not

It is not a defect in this PR. Recorded as gatehouse `FINDINGS.md` F-27: two of nucleus's ratcheted gates are at zero headroom, and four PRs were red simultaneously on gates working as designed. The clippy half of that is now fixed on this branch — your one cast is scoped — and this half is a repo-level decision about whether the floor keeps a margin.

Options, none of which are yours alone to take: cover ~392 more lines in the new code, exclude the harness from the measurement the way `nucleus-verifier-service` already is, or restore headroom to the floor deliberately.

PR #2768, opened 2026-09-10
fix(gatehouse): the prelude pin named a library that no longer exists

brandon-coproduct:
`.gatehouse/pipeline.writ` imports the `ci` prelude **by digest**, which is what makes the plan hermetic — and what makes the pin go stale in silence.

gatehouse's `cf535ce` added `imagePinsEnv_b` to `prelude/ci.writ`. That definition has no caller and enforces nothing, but importing it moved the prelude's digest, so our pin (`sha256:3180b330…`) named a library gatehouse no longer ships:

```
error at byte 751: no library for import "sha256:3180b3306a5a4ecd..."
```

`gate plan check` fails there, before elaborating anything.

**Nothing caught it.** `.github/workflows/gatehouse-plan.yml` is informational rather than a required context, and skips entirely unless `GATEHOUSE_TOKEN` is set. The last green run (2026-09-08, `nodes=5298`) predates the change, so no later run could red.

## Verified before bumping

Checked against gatehouse at `2476e955` that the plan is still **admissible**, not merely parseable:

```
admissible: proved by `admissible`
ok: 13 definition(s), 5 gate(s), plan hashed, admissibility proved;
    root `admissible` kernel-checked;
    nodes=5317 ctxs=1268 ty=4779 red=10780 sub=7579 shift=1027
```

So this is a stale pin, not a broken plan.

The rationale is recorded beside the import, because the next person hits this the same way: **an unwired predicate is not free**, and a hermetic pin with no updater behind an informational gate goes stale without a signal. Recorded as gatehouse `FINDINGS.md` F-21; the longer-term fix — something that fails when a pinned prelude stops existing — is not in this PR.

https://claude.ai/code/session_01MJ65muwqWqPidSYTnpknDi

### DIFF
--- a/.gatehouse/pipeline.writ
+++ b/.gatehouse/pipeline.writ
@@ -6,7 +6,15 @@
 --   Cap  = (net, exec, wallMs, cpuMs, memMb, fsRead, fsWrite, secrets)   net tag 3 0 = None, exec tag 3 1 = Scoped
 --   Gate = (name, hash, scope, env, cap, cmd, timeoutMs, outputs)         hash is filled by the elaborator (zeroed when hashed)
 -- env = the rust:1.96.1-slim-bookworm image index digest (rust-toolchain.toml pins 1.96.1).
-import "sha256:3180b3306a5a4ecd77485a0ea2e086df2a3ab0cb7bef20973bec2197c7d2c031" as ci
+-- The `ci` prelude is pinned by digest, which is what makes this plan hermetic -- and what makes
+-- the pin go stale in silence. gatehouse's cf535ce added `imagePinsEnv_b` to prelude/ci.writ;
+-- that definition has no caller and enforces nothing, but importing it moved the prelude's
+-- digest, so the previous pin (sha256:3180b330...) named a library gatehouse no longer ships and
+-- `gate plan check` errored "no library for import" before elaborating anything. Nothing red:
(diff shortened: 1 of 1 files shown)


brandon-coproduct:
Closing — there was no defect. I diagnosed this wrong.

The two pins were already **consistent**. `gatehouse-plan.yml` sets `GATEHOUSE_REF: 4d42510657690b90cf796d57ce277e665fca0989` and builds gatehouse at that commit; that commit's `prelude/ci.writ` hashes to exactly `sha256:3180b330…`, which is what this file imported. The plan check was passing and correct.

My error: I ran `gate plan check` against a gatehouse working tree on `main`, saw `no library for import`, and read a hermetic pin's refusal as staleness. A pin refusing a library the pinned build doesn't have is the pin *working*. Testing against the wrong side of a two-sided pin produced a finding that wasn't there — and this PR bumped one of the two coupled pins alone, which is what then made the informational check fail.

The real finding, which I've kept: **`GATEHOUSE_REF` and the prelude import digest must move together, and nothing checks that they agree.** This PR is the demonstration. That is a genuinely Tier 0 predicate — both operands are declarations (a workflow env var, an import line, and a file in a pinned repo) — so it is checkable with no toolchain and no tree of our own, and it would have refused this change at authoring time.

Recorded in gatehouse `FINDINGS.md` F-21, rewritten.

PR #2796, opened 2026-09-11
docs(adr-0007): make the defect unwritable — thirty-nine Rust mandates from the defect record

brandon-coproduct:
## What this is

A review read **252 defect records** from the full history of both repositories — 2,248 nucleus commits and 232 gatehouse commits plus 62 `FINDINGS.md` rows — scoring every fix and refactor against three architectures: writ matured as a plan language, a rewrite of the decision core in writ, and a proof-oriented Rust fork.

This ADR is the residue, and the residue is the largest actionable result:

> **58 of 252 records — 23% — were defects whose refusal needs no language extension, no verifier, no rewrite and no fork. The type discipline already existed in stable Rust and was not applied.**

The load-bearing case is `f7f9719b`. `DischargedBundle` was already `!Clone`, `!Copy` and `#[must_use]` — every affine signal Rust offers was present — and a one-shot authorization could still be replayed, because three signatures took it by `&`. Nothing was missing from the language. The refactor was the fix, and a written rule would have been cheaper. That is the thesis: where the mechanism exists and the defect recurs anyway, what is missing is a mandate, not a type system.

It is ADR 0006's finding one level down. 0006 found mechanisms built, machine-proved, and *not wired to the enforcement path*; this finds mechanisms built into the language itself and not reached for.

**Thirty-nine rules in nine families** — A sum types that lost a case · B defaults that grant · C evidence and witnesses · D order as a type · E records and exhaustiveness · F derive, never restate · G one decider per fact · H ambient authority · I gates that can fail. Each has a stable id, the commit whose defect motivated it, and a named enforcement tier, so that "review" reads as a gap rather than as coverage.

## The product test

**More confidence**, and peripheral to the frontier — which is the honest answer. This adds no authority dimension and moves neither ρ nor C(T). It turns a class of recurring defect into a written, cited, partly-mechanical rule, and it puts two gates on main that can fail.

## Invariants kept

No default widened. Nothing here touches authority, grants, ceilings or the lattice; the diff is one ADR, one `clippy.toml`, and two index/mandate pointers, with **591 insertions and 0 deletions**.

The two `clippy.toml` entries only ever *narrow* what compiles, and both are in service of `exercised authority ≼ delegated authority`: `transmute` fabricates a sealed witness without `unsafe` appearing at the mint site (C-1), and `Vec::leak` converts an owned must-use capability into a `&'static` that outlives every consumer (C-5).

## Validation

**A-19 was honoured for the two wired entries rather than asserted.** Red first, on Rust 1.96.1:

```
$ printf 'fn adr0007_probe(x: u32) -> i32 { unsafe { std::mem::transmute(x) } }\n' \
    >> crates/nucleus-net-probe/src/main.rs
$ cargo clippy -p nucleus-net-probe --all-targets
warning: use of a disallowed method `std::mem::transmute`
  = note: `#[warn(clippy::disallowed_methods)]` on by default

$ git checkout crates/nucleus-net-probe/src/main.rs
$ cargo clippy -p nucleus-net-probe --all-targets -- -D warnings
    Finished `dev` profile in 0.32s
```

That run also establishes **empirically that a root `clippy.toml` resolves for workspace members** — which the Clippy documentation does not state, and which `CARGO_MANIFEST_DIR` being the *member* directory gives reason to doubt.

Also run, all green:

- `cargo clippy -p portcullis-profiles -p nucleus-egress-probe -p nucleus-proto --all-targets -- -D warnings` → `Finished in 18.32s`, exit 0
- `bash ci/no-vendor-strings.sh` → `OK: vendor-neutrality CI gate clean`
- Every `adr/NNNN` link in `docs/SUMMARY.md` resolves to a file (7/7)
- Every commit sha cited in the ADR resolves — **this caught three defects in my own draft**: `0109d29`, `112bea6` and `ccc5c46` are *gatehouse* commits and were cited as if local, and `UNCOVERED_CEILING = 0` lives in gatehouse's `crates/xtask/src/gates.rs:21`, not nucleus's. All four now say so.

**Not yet run at time of opening:** full-workspace `cargo clippy --workspace --all-targets -- -D warnings`. It is in flight; I will post the result on this PR. The risk is bounded — `transmute` and `.leak()` are **0 occurrences** across `crates/`, `tools/` and `fuzz/` — but the claim is not made until the command has said so.

### Why only two entries are wired

CI runs `cargo clippy --all-targets --all-features -- -D warnings`, so one existing violation — in test code as much as in production — reds the merge queue. An entry is added only where the tree is already clean. Measured 2026-09-11:

| candidate | occurrences | wired |
|---|---|---|
| `std::mem::transmute` | **0** | yes |
| `std::vec::Vec::leak` | **0** | yes |
| `std::env::set_var` | 14, all `#[cfg(test)]`/comment | no — needs 14 `#[expect]` |
| `std::env::remove_var` | 13, all `#[cfg(test)]` | no — needs 13 `#[expect]` |
| `unwrap_or_default` | 218 | no — scoped dylint |
| `unwrap_or` | 584 | no — scoped dylint |
| `#[derive(Default)]` | 177 | no — scoped dylint |

H-1 (`no set_var`) is the next PR and is deliberately separate: **zero production call sites today**, so it is a real gate waiting on 27 uniform `#[expect]` annotations. Splitting keeps that diff mechanical and this one reviewable.

### Corrections to this analysis's own first pass

Recorded in the ADR because the wrong belief is worth more than the correction — the first pass counted imports and macro text as call sites:

| | first pass | measured |
|---|---|---|
| `Command::new` sites | 262 | **133** (64 were `std::process::exit`) |
| `libc::` / `nix::` sites | 91 | **53** |
| real `unsafe` blocks | 71 | **35**, in 10 files |

### What this does not claim

Stated in the ADR rather than left to be discovered: 58 of 252 is **not a majority**; these rules reach **no effectful defect directly** (103 records are in code that spawns, syscalls or awaits — C-4 and D-2 fence that surface, they do not enter it); **none addresses `unsafe` or UB** (35 blocks, none in the corpus); and only one of the three ways module privacy is re-opened — `transmute` — is expressible as a `clippy.toml` entry, so **C-1 is one third wired and says so**. `unsafe` field access and a derived `Deserialize` need the proposed dylint.

### Incidental

`docs/SUMMARY.md` also gains the **0006** entry, which was missing from the index.

https://claude.ai/code/session_01MUXVWHyPVkLspZcXE5gSZj

---
_Generated by [Claude Code](https://claude.ai/code/session_01MUXVWHyPVkLspZcXE5gSZj)_

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -144,6 +144,48 @@ forward-looking and the backlog is a migration, not a cleanup:
 A single command with no branching, no arithmetic, and no exit-code inspection. The moment there
 is an `if`, a count, or a pipeline whose status matters, it is Rust.
 
+## Mandate: make the defect unwritable
+
+**`docs/adr/0007-make-the-defect-unwritable.md` is thirty-nine rules, each derived from a defect
+this repository shipped and fixed. New Rust follows them; existing Rust follows them as each file
+is touched. Cite the rule id (`C-4`, `E-1`, …) in review.**
+
+They came out of a full read of both repositories' history: 252 defect records, of which **58 —
+23% — needed no language extension, no verifier and no rewrite.** The type discipline already
+existed in stable Rust and was not applied.
+
+The one to know by heart is **C-4**, because it is the reason the ADR exists. In `f7f9719b`,
+`DischargedBundle` was already `!Clone`, `!Copy` and `#[must_use]` — every affine signal Rust
+offers was present — and a one-shot authorization could still be replayed, because three
+signatures took it by `&`. Nothing was missing from the language. The refactor was the fix, and a
+written rule would have been cheaper.
+
+The nine families, in one line each:
+
+- **A — sum types that lost a case.** "Could not look" is never "looked and it was fine"; a `bool`
+  may not carry a three-valued decision; no blanket `map_err`.
+- **B — defaults that grant.** No `#[derive(Default)]` on a security type; `Option::None` may not
+  mean unrestricted; a `_ =>` arm denies.
+- **C — evidence and witnesses.** Evidence has a private constructor and is minted by the checker;
+  a capability is indexed by what it authorizes; a one-shot right is taken **by value**.
+- **D — order as a type.** Typestate, not source-line adjacency. A `compile_fail` doctest is not a
+  substitute for a type.
+- **E — records and exhaustiveness.** No `..` on a policy path — `E0027` is the mechanism.
+- **F — derive, never restate.** Derived serialization; one `Deserialize` struct and a `for` loop;
+  never a count restated beside the thing it counts.
+- **G — one decider per fact.** If a fact is written twice, delete one. A parity test leaves two
+  copies and converts the next drift into a test failure rather than an impossibility.
+- **H — ambient authority.** No `set_var`; an effect is reachable only from a witness.
+- **I — gates that can fail.** Every gate driven red on the real defect first (A-19).
+
+Three enforcement tiers — `clippy.toml`, a `tools/nucleus-*-lint/` dylint pass, or review — and
+the tier is named on every rule so "review" reads as a gap rather than as coverage. **A-19 binds
+the lints too: a lint ships only once it has been driven red on the commit its rule cites.**
+
+What the ADR does **not** claim: 58 of 252 is not a majority, these rules reach no effectful
+defect directly (103 records are in code that spawns, syscalls or awaits), and none of them
+addresses `unsafe`. Those limits are in the ADR rather than left to be discovered.
+
 ## Startup Loops
 
 On init, read `LOOPS.md` (git-ignored, local only) and start any loops defined there.
--- a/clippy.toml
+++ b/clippy.toml
@@ -1,0 +1,32 @@
+# Clippy configuration for the workspace.
+#
+# Entries here back the mandates in `docs/adr/0007-make-the-defect-unwritable.md`.
+# Every rule id below (C-1, C-5, ...) dereferences to a rule in that ADR, and every
+# rule cites the commit whose defect motivated it.
+#
+# A-19 applies: an entry is added only once it has been driven RED on a real
+# violation and green with the violation removed. An entry that has never fired is
+# a claim, not a gate. `docs/adr/0007` §Enforcement records how each was probed.
+#
+# ENTRIES ARE ADDED ONLY WHEN THE TREE IS ALREADY CLEAN OF THEM. `cargo clippy`
+# runs `--all-targets -D warnings` in CI, so a single existing violation — in test
+# code as much as in production — reds the merge queue. Where a mandate is not yet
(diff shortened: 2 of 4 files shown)


brandon-coproduct:
## Full-workspace clippy result (the run promised in the PR body)

**Green, and the two new `clippy.toml` entries are clean across the whole tree.**

```
$ cargo clippy --workspace --exclude nucleus-verifier-service \
      --all-targets --all-features -- -D warnings
    Finished `dev` profile in 1m 43s
exit 0
```

`clippy::disallowed_methods` fired **0 times across 313 crates**, in two independent feature configurations (default and `--all-features`). That was the only thing at risk from this diff, and it is settled.

### Why one crate is excluded

`nucleus-verifier-service` cannot be clippy'd under `--all-features` in a fresh clone, and this is documented behaviour rather than a regression. `--all-features` turns on `embedded-wasm`, and `routes.rs:55` does `include_str!("../../../sdks/verifier-js/pkg/nucleus_verifier_wasm.js")` against a path that `sdks/verifier-js/.gitignore` excludes. The crate's own `Cargo.toml` says so:

> CI and release builds turn it on, having built the SDK first; with it off the two `/static/wasm/*` routes answer 503 and name the command.

I did not run `wasm-pack build` first. Nothing in this PR touches that crate.

---

## Two pre-existing findings surfaced by the validation

Neither is caused by this diff — it is 4 files, zero source — and neither is fixed here. Filing them because the repository's rule is that a finding is recorded in the change that found it.

### 1. `portcullis-core` lib tests do not compile under default features

```
error[E0433]: cannot find module or crate `serde_json`
   --> crates/portcullis-core/src/prov_export.rs:413:20
```

`serde_json` is `optional = true`, reachable only via `artifact` or `zkvm`. The test `serialized_document_omits_relation_keys_entirely` uses it and carries no `#[cfg(feature = ...)]`. It passes under `--all-features` (verified, `Finished in 1m 13s`), which is why CI is green — **CI runs `--all-features` and never exercises the default-feature build of that target.**

This is a third instance of the shape `d73e874c` fixed ("gate the two feature-dependent test targets", #2732), and an instance of the family ADR 0007 names: *the program that is checked is not the program that ships*. One-line fix; happy to send it separately.

### 2. A clippy invocation weaker than CI's fails, and nothing says so

Both findings above have the same root: `cargo clippy -p <crate> --all-targets` — the natural local command — fails on two crates for reasons unrelated to the code being reviewed, because the tree assumes `--all-features` plus a prebuilt JS SDK. There is no `CONTRIBUTING.md` and nothing names the prerequisite. That is `90e3c18a`'s shape ("the clippy gate declares what it actually reads") on the developer-facing side rather than the CI side.

---

**Summary for review:** the diff is documentation plus two lint entries that are provably clean workspace-wide and were driven red-then-green before being added. The two findings above are pre-existing, invisible to CI by construction, and scoped out of this PR deliberately.


---

PR #2825, opened 2026-09-11
gate: a gate's own timeout must fire before its job is killed

brandon-coproduct:
Three open PRs all showed the same red check. It was **one root cause, it was on main, and no PR had caused it** — the shadow lane's clippy entry had not been failing. It had not been running.

## What was wrong

Two clocks ran over the same work, and nothing compared them:

| clock | who enforces it | does it report? |
|---|---|---|
| the gatehouse action's `timeout` | the runner | **yes** — a verdict |
| the job's `timeout-minutes` | GitHub | no — a `cancelled` conclusion |

The clippy entry declared `timeout: "2700"` inside a `timeout-minutes: 45` job — the same 2700 seconds — and `compare` defaults to `true`, so the command runs a **second** time to produce the agreement line that is the whole point of shadow mode. The declared need was 5460s inside a 2700s job. Even one run did not fit: setup spends 43s before the gate step starts (measured, run 34625269465).

So GitHub was always first, and every overrun came back `cancelled` — the same conclusion an infrastructure fault gives. The one mechanism that would have said *"this gate exceeded its budget"* was preempted by the one that says nothing at all.

## What it cost

Measured, not estimated:

- Last green run **2026-09-09T11:46:49Z**. Between then and 2026-09-11T18:00Z, **304 runs produced no verdict** (252 `cancelled`, 52 `failure`).
- **14 of 20** sampled cancelled runs had spent the **full 45 minutes** to produce nothing; the other 6 were superseded by `cancel-in-progress`.
- The earlier 52 were a different, now-resolved cause: `curl: (22) ... 404` → *"the control plane refused this run's token"*. Today's runs get past trust and hang in the command.
- Sibling matrix entries are unaffected and prove the parser sees a real difference: `fmt` and `ci-spec` finish in **1 minute**.

## The gate

`cargo xtask gate-budget`, decided from declarations alone — the workflow and the action definition. No source tree, no toolchain, no network, same verdict against an empty checkout. For each step using the action:

```
runs * timeout_s + setup  <  timeout-minutes * 60
```

strictly, with `runs = 2` when `compare` is on. Equality is the defect above, not a boundary case that just fits.

Both defaults are **read from `.github/actions/gatehouse/action.yml`** rather than copied into the gate. A second copy of a default is a second thing to drift, which is the failure this family exists to refuse — `fly_pools.rs` calls the manager's own validator for the same reason. It also reds if a workflow starts using the action without being listed, and if the parser stops finding any step at all: a gate that measures nothing passes everything.

## The fix, stated narrowly

`timeout: "1200"` — 1200 x2 + setup = 2460s, inside the job's 2700s, so the **runner** reports first.

This does **not** claim clippy now passes. It has never once completed, so nobody knows what it costs. It claims the opposite is no longer possible: an overrun now produces a timeout verdict at 20 minutes instead of a silent cancellation at 45 — and costs less runner time while doing it. **The number it reports is the measurement that was missing.** If 20 minutes turns out to be too little, that is a finding this lane can finally produce.

## Verification

- Driven **RED on the live defect** before green (A-19): `EXIT=1` on the shipped 2700/45 declaration, `EXIT=0` after the fix.
- Probed in `scripts/check-gates-can-fail.sh`; verified red→restore→green with the tree clean afterwards, and `timeout-minutes` untouched.
- `cargo test -p xtask` RC=0 (8 new tests), `cargo fmt --all -- --check` RC=0, `cargo clippy -p xtask --all-targets -- -D warnings` RC=0.

One parser bug was caught during development and is worth recording: the gate first reported the action's *default* 3600s over a step that plainly declared 2700. `with:` is a **sibling** of `uses:`, not a child, so the block scan ended before reading either input. It was caught only because the printed number disagreed with the declaration read by eye — a gate reporting a plausible wrong number is the failure mode this whole family is built against.

### DIFF
--- a/.github/actions/gatehouse/gatehouse.sh
+++ b/.github/actions/gatehouse/gatehouse.sh
@@ -118,8 +118,27 @@ say "receipt $INDEX in $(jq -r .log.origin <<<"$EXCHANGE"): $VOUT"
 # ── 6. The plain run, for the agreement line ───────────────────────────────────────────────
 AGREE=skipped; GITHUB=unknown
 if [ "$GH_COMPARE" = true ]; then
-  set +e; bash -c "$GH_RUN" >/dev/null 2>&1; FC=$?; set -e
+  # Bounded by the SAME budget as the gatehouse run. `cargo xtask gate-budget` computes the
+  # job's requirement as `runs * timeout + setup`, and with compare on `runs` is 2 -- which
+  # is only true if this second run is bounded too. It was not, so the arithmetic that gate
+  # checks described a bound that did not exist on this side.
+  #
+  # No slack here, unlike the runner above. The slack there exists so the runner can hit its
+  # own deadline and write an honest `errored` receipt first; this is a plain `bash -c` with
+  # no receipt to write and nothing to wait for, so the budget is the budget. The worst case
+  # across both runs is therefore (timeout + 120) + timeout, and that 120 is a term in
+  # `gate-budget`'s arithmetic (OUTER_SLACK_S) rather than something setup is assumed to
+  # absorb -- an assumed absorption is how the declared budget came to describe a bound that
+  # did not exist.
+  #
+  # A timed-out plain run is `fail` rather than fatal, deliberately: this half exists to say
+  # what GitHub's own check would have said, and "it did not finish in the budget" is a
+  # legitimate answer to that. Disagreement is never fatal in shadow mode.
+  set +e; timeout --kill-after=30s "$GH_TIMEOUT" bash -c "$GH_RUN" >/dev/null 2>&1; FC=$?; set -e
   if [ "$FC" -eq 0 ]; then GITHUB=pass; else GITHUB=fail; fi
+  if [ "$FC" -eq 124 ] || [ "$FC" -eq 137 ]; then
+    say "the plain run exceeded the same ${GH_TIMEOUT}s budget; counted as fail for the agreement line"
+  fi
   if { [ "$HELD" = held ] && [ "$GITHUB" = pass ]; } || { [ "$HELD" = not-held ] && [ "$GITHUB" = fail ]; }; then AGREE=true; else AGREE=false; fi
 fi
 
--- a/.github/workflows/gatehouse-shadow.yml
+++ b/.github/workflows/gatehouse-shadow.yml
@@ -131,7 +131,18 @@ jobs:
           # runner (gatehouse F-100) did exactly that on sixteen open pull
           # requests. A gate's own timeout must fire before its job is killed,
           # or the gate cannot report its own failure.
-          timeout: "2100"
+          #
+          # 2100 -> 1200, and the reasoning above is unchanged: this is the same rule applied
+          # to BOTH runs. `compare` defaults true, so the command runs a second time for the
+          # agreement line, and this PR bounds that run too (gatehouse.sh) — so the job must
+          # hold 2 x timeout + setup, not one. At 2100 that is 4260s inside a 2700s job, which
+          # fails the same way 2700 did, one run later. 1200 x2 + 60 = 2460s fits.
+          #
+          # `cargo xtask gate-budget` decides this from the declarations, and reds on 2100.
+          # If the agreement line is worth more than the second run's deadline, the other
+          # answer is `compare: false` for this entry and 2100 stands — that is a judgement
+          # about what shadow mode is for, and it belongs to whoever set 2100.
+          timeout: "1200"
           bin-dir: gatehouse/target/debug
           control-url: ${{ vars.GATEHOUSE_CONTROL_URL || 'https://gatehouse-controld.fly.dev' }}
       - name: The agreement line
--- a/.github/workflows/gatehouse-shadow.yml
+++ b/.github/workflows/gatehouse-shadow.yml
@@ -132,16 +132,23 @@ jobs:
           # requests. A gate's own timeout must fire before its job is killed,
           # or the gate cannot report its own failure.
           #
-          # 2100 -> 1200, and the reasoning above is unchanged: this is the same rule applied
-          # to BOTH runs. `compare` defaults true, so the command runs a second time for the
-          # agreement line, and this PR bounds that run too (gatehouse.sh) — so the job must
-          # hold 2 x timeout + setup, not one. At 2100 that is 4260s inside a 2700s job, which
-          # fails the same way 2700 did, one run later. 1200 x2 + 60 = 2460s fits.
+          # Fitting the declaration is necessary and was NOT sufficient, and I claimed it would
+          # be. Run 34637222980 carried this 1200 and the gate step still ran 44m35s and was
+          # still killed by the job's cap with no verdict: `duration_ms=2675520` against
+          # `2676029` for the original 2700 run — half a second apart. The runner does not
+          # enforce its own timeout, so no value declared HERE makes it report. That is why
+          # the budget is also enforced by the caller in `.github/actions/gatehouse` — the
+          # declaration says what the bound is, the action is what makes it bite.
           #
-          # `cargo xtask gate-budget` decides this from the declarations, and reds on 2100.
+          # 2100 -> 1200 for arithmetic the declarations already contain: `compare` defaults
+          # true, so the command runs a second time for the agreement line, and the job must
+          # hold 2 x timeout + setup rather than one. At 2100 that is 4260s inside a 2700s
+          # job, which fails exactly the way 2700 did, one run later. 1200 x2 + 60 = 2460s
+          # fits. `cargo xtask gate-budget` decides this and reds on 2100.
+          #
           # If the agreement line is worth more than the second run's deadline, the other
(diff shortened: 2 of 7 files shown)


brandon-coproduct:
**Verification status of one claim in this PR — still open.**

The PR says an overrun now "produces a timeout verdict at 20 minutes instead of a silent cancellation at 45". **That half is not yet verified**, and I would rather flag it than let it merge as though it were.

What is confirmed: `GH_TIMEOUT: 1200` reaches the runner on this branch, so the declaration change is live.

What is not: on run `34634762509` the gate step ran **25m31s** (18:43:26Z → 19:08:57Z) with that 1200s timeout and produced **no verdict line** before being cancelled. I cannot yet distinguish between

1. the runner timing out silently at 20 minutes and the remaining ~5½ minutes being the plain `compare` run, which is the intended behaviour; and
2. the runner not enforcing its declared timeout at all — in which case this fix is **necessary but not sufficient**, since no value of `timeout` would make it report first.

Both attempts to measure it were destroyed by my own pushes to this branch: `gatehouse-shadow.yml` sets `cancel-in-progress` for `pull_request`, so each push killed the in-flight run at ~26 minutes, twice, ~15 minutes short of the ~41 it needs.

**The gate itself is unaffected either way.** It compares two declarations and is correct on its own terms; a job that cannot let its runner report first is a real defect whether or not the runner would then report. Only the *repair's* effectiveness is in question.

Next step is simply to leave this branch alone for ~45 minutes and read the result.

brandon-coproduct:
**Resolved — and it refutes the claim I flagged. Corrected in `122b3801`.**

The run finally completed a full cycle without my pushes killing it:

| run | declared `timeout` | gate step | outcome |
|---|---|---|---|
| `34625269465` (original defect) | 2700 | `duration_ms=2676029` | cancelled, no verdict |
| `34637222980` (with this fix) | **1200** | `duration_ms=2675520` | cancelled, no verdict |

**Half a second apart.** `GH_TIMEOUT: 1200` is confirmed reaching the runner in the second run's own log, and the runner still produced nothing after its tenant line for **44m35s** before the job's 45-minute cap killed it. In 998 log lines the only occurrences of "timeout" are the input echo and the env echo.

So outcome (2) from my earlier comment holds: **the runner does not enforce its declared timeout, and no value here makes it report first.** That is a defect one level below this PR — in gatehouse's runner, not in nucleus's declarations. The PR description's claim that an overrun would "produce a timeout verdict at 20 minutes" was wrong; three claim sites are now corrected in the tree (the shadow workflow comment, the manifest-guards comment, and the module header).

**The gate is unchanged and still right, and the distinction is why it's worth keeping.** It decides whether the *declarations* are coherent, and a budget that cannot fit inside the job running it is incoherent whether or not the runner would have honoured it. What must not be claimed is that fixing the arithmetic makes an overrun report.

`timeout` stays at **1200** rather than reverting to 2700: a declaration that cannot fit is wrong independently of the runner bug, and when the runner is fixed, 1200 is already the value that lets it speak.

One process note worth recording: this measurement took three attempts because `gatehouse-shadow.yml` sets `cancel-in-progress` for `pull_request`, so each of my own pushes killed the in-flight run at ~26 minutes — about 15 short of the ~41 it needed. Holding the branch still for one full cycle is what produced the number that refuted my own claim.

brandon-coproduct:
**Verified. The enforcement works, and the claim this PR originally made is now true — by a different mechanism than it claimed.**

Run `34643845546`, clippy shadow gate step:

| | before | after |
|---|---|---|
| gate step duration | 44m 35s | **20m 00s** — exactly the declared 1200s |
| conclusion | `cancelled` | **`failure`** |
| verdict in log | none | yes |

```
##[error]gatehouse: the gate exceeded its declared 1200s budget and was stopped.
That is a VERDICT about this gate, not an infrastructure fault: the command did not
finish in the time its declaration allows. Raise the action's `timeout` (and the job's
`timeout-minutes` with it — `cargo xtask gate-budget` checks that they fit), or make
the gate faster.
```

Two zero-second boundaries worth noting: 20:24:02 → 20:44:02 is 1200s to the second, and the conclusion changed from `cancelled` (an infrastructure-shaped non-answer) to `failure` (a verdict about the gate).

**What was actually wrong, restated for the record.** Lowering `timeout` 2700 → 1200 changed nothing — measured, `duration_ms` 2676029 vs 2675520, half a second apart. The runner does not enforce the budget it is handed. What fixed it was **bounding the command in the caller** (`timeout --kill-after=30s "$GH_TIMEOUT"` around both the runner and the plain compare run), which is a different claim from the one the PR opened with and is the one now supported by evidence.

That distinction is the whole content of gatehouse F-86 and F-87: *a timeout that travels as data to a component you do not control is a request, not a limit.*

The gate itself was never in question — it decides whether the declarations are coherent, which they were not, independently of whether the runner honoured them.

brandon-coproduct:
Dequeued deliberately, so #2856 lands first. Explaining rather than silently reshuffling.

**They conflict.** Test-merged `main` + this + #2856: `rc=1` on both shared files, `.github/actions/gatehouse/gatehouse.sh` and `.github/workflows/gatehouse-shadow.yml`. Both were queued (this at 2, #2856 at 4), so whichever landed second would have been ejected — and the ejection would have fallen on the better change.

**#2856 is better on both files, and not marginally:**

- Its outer deadline is `GH_TIMEOUT + 120`, with slack **so the runner can hit its own deadline, kill the command and write an honest `errored` receipt**. This PR bounds at exactly `GH_TIMEOUT`, which races that and can kill the runner in the moment it was about to report. That is a correctness bug in mine.
- It renames the check so a red reads *"could not look, never disagreement"*. Sixteen red shadows read as sixteen disagreements when every one meant the loop could not run. I looked at that name for six loops without questioning it.

**What this PR should keep is the gate**, which #2856 does not have: `cargo xtask gate-budget`, its untimed-jobs ratchet, and the probe. When #2856 lands I will rebase this onto it, drop my two overlapping files, and keep the gate.

**One thing to settle first, because the gate will red on main otherwise.** `gate-budget` checks `runs × timeout + setup < timeout-minutes × 60`, and `compare` defaults true:

```
#2856's 2100 x2 + 60 = 4260s  >  2700s job   → FAIL
```

That is the same hole #2856 closes for the first run, still open for the second. Three ways out, on the PR: `timeout: 1200`; `compare: false` for clippy; or bound the compare run as well — `gatehouse.sh:121` is still a bare `bash -c "$GH_RUN"`, and F-100's deadlock class applies to it identically. The third is the one that makes the `runs × timeout` arithmetic true rather than optimistic, and it is the one thing this PR has that #2856 lacks.

brandon-coproduct:
Rebased onto main (which now carries #2856). Force-pushed `82ad429`.

**What I dropped.** My `gatehouse.sh` hunk. #2856's outer deadline is the better version — `timeout --kill-after=30 $(( GH_TIMEOUT + 120 ))` gives the runner room to hit its own deadline and write an honest `errored` receipt before the outer one fires, which mine did not. That side of this PR is now just the delta over it.

**What is left in `gatehouse.sh`.** The plain `compare` run. Line 121 on main is still `set +e; bash -c "$GH_RUN"` with nothing around it, so a gate that hangs the way F-100 describes hangs there instead, one step later, and the job is cancelled exactly as before. Now bounded at `$GH_TIMEOUT`, no slack — a plain `bash -c` has no receipt to write, so there is nothing to wait for. A timed-out plain run counts as `fail` for the agreement line rather than being fatal.

**The budget, which overrides your 2100.** Two corrections, both from arithmetic the declarations already contain:

1. `compare` defaults true, so the command runs **twice**. The job must hold `2 x timeout`, not one. At 2100 that is 4260s inside a 2700s job — the same shape as 2700, one run later.
2. The action bounds the runner at `timeout + 120`, not `timeout`. That slack is real wall time the job has to hold, so it is now a term (`OUTER_SLACK_S`) rather than something `SETUP_ALLOWANCE_S` was quietly assumed to absorb. A budget that describes a tighter bound than the one actually enforced is this gate's own defect, one level down.

So the rule is `runs * timeout + 120 + 60 < timeout-minutes * 60`, and at 1200 that is **2580s inside 2700s**. Green:

```
ok   .github/workflows/gatehouse-shadow.yml:41  job 2700s, gate 1200s x2 + 120s slack + 60s setup = 2580s
ok   23 job(s) without a timeout, all pinned
```

`the_outer_slack_is_what_decides_this_site` is the falsifier for the second correction: 1300s twice fits a 2700s job by 40 seconds without the term and overruns by 80 with it.

**If you want 2100 back**, the other exit is `compare: false` for this entry — 2100 + 120 + 60 = 2280 fits fine with one run. That trades the agreement line for the larger budget, which is a judgement about what shadow mode is for and yours to make. Say the word and I will switch it; I picked 1200 because the agreement line is the thing shadow mode exists to produce.

Pre-push, exit codes not tails: `actionlint` 0, `shellcheck -S error` 0, `check-ci-spec.sh` 0, `cargo test -p xtask` 0 (163 passed), `gate-budget` 0, `check-gates-can-fail.sh` 0 in a clean worktree (40 probed, ceiling 6, 0 unaccounted).

PR #2829, opened 2026-09-11
A decision authorises one operation, not a session of them

brandon-coproduct:
Stacked on #2827. **linearity 35 → 4**, and the metric from that PR is what proves it.

## The defect

`DecisionToken` carries every affine signal Rust offers, and its own doc says so:

> - **Linear**: non-Clone, non-Copy — cannot be reused
> - **`#[must_use]`**: compiler warns if dropped without consumption
> - **Sealed**: private `_seal` field prevents external construction
>
> Only `Kernel::decide()` can create this token. Kani proof `proof_decision_token_unforgeable` verifies no other construction path exists.

**Thirty-one signatures took it by `&`.**

So one `Kernel::decide()` authorised an unbounded number of operations. The scope check held — `require_decision_for` refuses a token about a *different* operation, in every profile — so a `ReadFiles` decision could not pay for a write. It could pay for **every read**.

That is `f7f9719b` again, the record ADR 0007 C-4 is built on, at thirty-one times the scale.

## A correction to ADR 0006 C2.5

Its line about this is **stale**: *"`DecisionToken` is inert outside `debug_assert` (item 25)"*. It is not inert. `#2799` — *"the decision-scope check is enforced in release, and thirty copies become one"* — made it real.

What was left was purely the calling convention, which is a different defect from the one the ADR names. Worth fixing on the record, since C2.5 is the row that owns this work.

## The compiler found every caller

None were guessed. Two rounds of driving the fix off `error[E0308]` `line:column` pairs.

An early attempt used a text regex instead and stripped `&token` from an unrelated `&str` call across 28 files — reverted, and redone from compiler output only. Worth noting because it's the same lesson as the lint that matched paths by text: **a matcher that can silently hit the wrong thing is worse than one that can't run.**

## The test that was depending on the defect

The only judgement call. One test reused a single token across four opens, and `use of moved value: rt` caught it the moment the type stopped allowing it:

```rust
let rt = token(&mut kernel, Operation::ReadFiles, "seed.txt");
refused!("open",           sandbox.open("seed.txt", rt, wrong()));
refused!("read",           sandbox.read("seed.txt", rt, wrong()));
refused!("read_to_string", sandbox.read_to_string("seed.txt", rt, wrong()));
refused!("open_dir",       sandbox.open_dir("seeddir", rt, wrong()));
```

It now mints one per call through a single closure — three would each need `&mut kernel` at once.

## The measurement

```
linearity   35 → 4     ceiling lowered to 4
```

The four that remain are all `DischargedBundle`: `ifc_api.rs`'s `_proof`, the audit-log renderer `discharge_witness`, and two effect signatures. That's the `f7f9719b` shape at its own origin — the next change, not this one.

## Verification

`cargo check --workspace --all-targets --all-features` clean (bar the pre-existing `nucleus-verifier-service` wasm-pack artifact, which needs `wasm-pack build` and which `.clippy-ratchet.toml` already records) · clippy `-D warnings` clean · `cargo test -p nucleus --all-features` **81 passed, 0 failed** · `cargo fmt --all --check` clean · `check-gates-can-fail.sh` 36 probed, 0 unaccounted.


brandon-coproduct:
`Detect changed crates` here is the same class as #2843, not a defect in your diff:

```
fatal: couldn't find remote ref adr/convergence-ratchet
##[error]Process completed with exit code 128.
```

This PR's base was `adr/convergence-ratchet` (#2827). That merged and the branch was deleted; GitHub has retargeted this PR to `main`, but the in-flight run still carries the old ref in its event payload, so every job that resolves `origin/$BASE_REF` dies with git 128.

**A rerun will not clear it** — I tested that on #2843. A rerun replays the original event, so it faithfully reproduces the stale ref. Only a new trigger picks up the retarget, which means the rebase this PR needs anyway for being `DIRTY`.

Third instance in two days of the same shape, all in this stack — recorded as gatehouse **F-147**. The cost is that the failure text names a git ref and the job is called "Detect changed crates", so nothing in between points at a merged parent.

PR #2835, opened 2026-09-11
tot: how many crates have declared themselves panic-free

brandon-coproduct:
The third family on the card, and the first whose number is genuinely low. The badge now names it.

```
  scorecard | tot 16.25%

  family  unit                              population  discharged         %  undeclared
  bound   witness-accepting parameter site         172         171    99.41%          32
  alg     (lattice type, law) obligation           278         110    39.56%           8
  tot     production crate                          80          13    16.25%           5
```

Three commits, in dependency order.

## 1 — a proof harness, a build script and a `_test.rs` are not production

`is_production_path` defines the corpus five gates range over and let through three kinds of file that ship in no binary: three Kani harnesses, `crates/nucleus-econ-kernels/proofs/`, `certificate_convergence_test.rs` (`_test.rs` — the original rule said `_tests.rs` and missed by one character), and two `build.rs`.

The Kani case is worse than an omission: a proof harness is written against a deliberately narrowed model — `portcullis` marks whole fields `#[cfg(not(kani))]` — so counting it measures the proof rather than the binary. The coverage gate already excludes it as `--ignore-filename-regex '(tests/|kani\.rs|main\.rs)'`.

**Measured on the family this unblocks:** `.unwrap()`/`.expect()` in production falls 440 → 393, and **`portcullis` drops 89 → 43 and stops being the worst crate** — `nucleus-tool-proxy` at 81 is. A gate reading the loose corpus reports roughly twice the truth for `portcullis` and names the wrong crate to fix first.

**No pinned ratchet moves.** The corpus goes 691 → 684 files and `law-mechanisms` (9 dead, 132 allowances), `inert-authority` (33), `bound` (171/172), `convergence` and the scorecard all report exactly what they did before. That is why it is its own commit: the claim worth checking is that the corpus changed and no verdict did.

A test pins that the exclusions are segments and whole filenames, not substrings — `kani-utils`, `proofs_index.rs`, `kani_support.rs` and `testing.rs` all still ship.

## 2 — thirteen crates declare themselves panic-free

None of the seven lints that decide totality was configured anywhere in this repo. The production region carries **1,787 sites**:

| lint | sites |
|---|---|
| `arithmetic_side_effects` | 912 |
| `indexing_slicing` | 616 |
| `expect_used` | 155 |
| `unwrap_used` | 87 |
| `panic` | 10 |
| `unreachable` | 7 |

That is a program, not a gate, so this follows `clippy.toml`'s own rule — *"entries are added only when the tree is already clean of them"* — and annotates the thirteen crates that measure zero.

**`cfg_attr(not(test), …)` and not a bare `deny`, because `assert!` is a panic.** Of the thirteen crates clean over lib and bins, only **two** are also clean under `--all-targets`; the other eleven are clean where it matters and panic in their own test modules, which is what tests are made of. The declaration is scoped to the shipped build — the same line `production_region` draws when it strips the test region.

## 3 — the `tot` family

The gate does **not** recompute the 1,787-site census: it takes minutes and this runs in the fast `Manifest Guards` job. It asks the binding question instead, which is a grep — *how many crates have declared themselves panic-free*. A count of sites nothing enforces is a measurement; a crate-level deny is a gate. The census is recorded in the ratchet with the command that reproduces it.

Two things the census refuses to fudge:

- **All seven lints or none.** A crate denying six can still panic, and crediting it would make the ratio mean "mostly".
- **Unmeasurable is not clean.** Eighteen crates showed zero; **five showed zero because clippy never looked** — four workspace `exclude`s (wasm, maturin, RISC-V) and `nucleus-verifier-service`, which needs a wasm artifact CI builds. Counting them as discharged is ADR 0007 A-2 exactly: *"I could not look" is never "I looked and it was fine."* They are reported as `undeclared`, the population is the 80 crates clippy can see, and a crate on that list that *does* carry the attribute is still not credited — so annotating something nothing compiles cannot raise the ratio.

## Verification

| perturbation | result |
|---|---|
| `v[0] + 1` appended to `nucleus-pca` | RED on both `indexing_slicing` and `arithmetic_side_effects` |
| one lint dropped from a crate's declaration | RED — `tot` 16.25% → 15.00% |

Gate-of-gates: 41 probed, 0 unaccounted, exit 0 — the scorecard now reds on three independent subjects, one per family. All thirteen annotated crates pass `cargo clippy --all-targets --all-features -- -D warnings`. 206 xtask unit tests pass, 11 of them new.

### DIFF
--- a/.scorecard-ratchet.toml
+++ b/.scorecard-ratchet.toml
@@ -142,8 +142,8 @@ population_floor = 278
 # 67 remaining crates need their sites fixed before they can declare; the two
 # tractable lints there are unwrap_used (87) and panic (10), and the 1528
 # arithmetic and indexing sites are a program, not a gate.
-floor_bp = 1625
-population_floor = 80
+floor_bp = 1728
+population_floor = 81
 
 [family.life]
 # 7 affine rights, 0 with a validity interval.
--- a/.scorecard-ratchet.toml
+++ b/.scorecard-ratchet.toml
@@ -88,3 +88,59 @@ population_floor = 172
 # lands.
 floor_bp = 3956
 population_floor = 278
+
+[family.tot]
+# 13 of 80 production crates declare themselves panic-free for the shipped
+# build. A function whose signature says `-> T` and panics is lying about its
+# type: #1203 (FlowTracker::label(0), u64 underflow on an unguarded sentinel),
+# #1184 (truncate() on multi-byte UTF-8), #481 (Ed25519 keygen .expect() in a
+# hot path), #2395 (a guest_cid other than 3 panics the guest, silently).
+#
+# THE SITE CENSUS, recorded here rather than gated, per clippy.toml's own rule:
+# "Where a mandate is not yet clean, the ADR records the measured count and the
+# ratchet instead of putting a red gate on main."
+#
+#   2026-09-11, production region, 1787 sites:
+#     arithmetic_side_effects  912
+#     indexing_slicing         616
+#     expect_used              155
+#     unwrap_used               87
+#     panic                     10
+#     unreachable                7
+#     todo                       0
+#
+#   cargo clippy --workspace --exclude nucleus-verifier-service --lib --bins \
+#     --all-features --message-format=json -- \
+#     -W clippy::unwrap_used -W clippy::expect_used -W clippy::indexing_slicing \
+#     -W clippy::arithmetic_side_effects -W clippy::panic \
+#     -W clippy::unreachable -W clippy::todo -A clippy::all
+#
+# That takes minutes and this gate runs in the fast Manifest Guards job, so the
+# gate does not recompute it. It asks the binding question, which is a grep: how
+# many crates have DECLARED themselves panic-free. A count of sites nothing
+# enforces is a measurement; a crate-level deny is a gate.
+#
+# WHY cfg_attr(not(test), deny(...)). assert! is a panic. Of the 13 crates
+# measuring zero over lib and bins, only 2 are also clean under --all-targets --
+# the other 11 are clean where it matters and panic in their own test modules,
+# correctly. So the declaration is scoped to the shipped build, the same line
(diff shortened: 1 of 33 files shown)


brandon-coproduct:
The `Gates must fail on their own subject` red here is **not a broken probe**, and the message it prints is misleading. Both reds on this PR are one cause.

The suite reports:

```
FAIL  xtask scorecard — still failing (exit 1) after restore      x4
```

That reads as a perturbation leaving something behind. It is not. `xtask scorecard` is **already red on this branch before any perturbation** — the same red as `Manifest Guards` two checks away:

```
tot fell to 27.16% against a floor of 27.50%: an obligation the tree declares stopped being discharged.
```

`probe_xtask` measured the gate perturbed and then restored, and read a red restore as *"the perturbation broke it"*. On a tree where the gate is already red it can only ever say that, four times, naming the one thing that did not happen. #2870 measures the baseline first, so an already-red gate says so and stops — a gate failing for its own reasons decides nothing under perturbation.

So: **one root cause, two symptoms.** The scorecard floor is the thing to settle. Per gatehouse F-105 it fails identically on branches with different diffs, which makes it a property of the base rather than of this PR — whether that means lowering the floor or discharging the obligation is yours, but it is not work this branch introduced.

brandon-coproduct:
Same cause as #2838 — full arithmetic there: https://github.com/coproduct-opensource/nucleus/pull/2838#issuecomment-5643371289

Short version: `[family.tot] floor_bp` on this branch is exactly *d*/**80**, and the tree measures exactly *d*/**81**. `discharged` is unchanged on both sides — the population grew by one crate, so the ratio fell at a constant numerator. The gate says *"an obligation the tree declares stopped being discharged"*, which sends you looking for a removed `#![deny]` that is not there.

Here it is `1625 = 13/80` pinned against `1604 = 13/81` measured.

`Gates must fail on their own subject` is a consequence of the same red, misreported as a restore failure; #2870 fixes that message.

PR #2838, opened 2026-09-11
PermissionLattice satisfied none of the 15 laws it declares

brandon-coproduct:
Found by running the `alg` family's own checker over the type it scores worst — 15 obligations declared, 1 discharged.

## What was wrong

`PermissionLattice` derived `PartialEq`, so equality compared `id`, `description` and `derived_from` — audit provenance that `meet`, `join` and `leq` all deliberately ignore. And `meet` mints a fresh UUID on every call:

```rust
id: Uuid::new_v4(),
description: format!("meet({}, {})", self.description, other.description),
derived_from: Some(self.id),
```

`verify_lattice_laws` — the checker that has sat in `portcullis-core` all along — over three of the type's own constructors: **99 violations.**

- `a.meet(&a) != a`
- `a.meet(&b) != b.meet(&a)`
- `a.meet(&b) == a` was **never** true for any operands, so `a ≤ b ⟺ a ∧ b = a` was broken by construction rather than by any particular pair

It also meant two policy-identical permission sets compared unequal, which is the wrong answer to the only question `==` is ever asked here.

## The fix is the one the tree already makes next door

`portcullis_core::attenuation::LiteralDelegation` hand-writes `PartialEq` because *"deriving `PartialEq` would break join commutativity: `a ∨ b` and `b ∨ a` list elements in different orders."* Same wall, same answer — make `PartialEq` **be** the law equality rather than adding a second notion beside it.

**99 → 3.** Commutativity, associativity, idempotence and leq/meet consistency now hold, and five tests pin them.

Nothing depends on the old behaviour: no production or test site compares two `PermissionLattice` values. I checked before changing it.

## The remaining three are not an equality bug

Absorption. `a ∧ (a ∨ b) = a` is precisely the law a **closure operator** breaks, and `meet` applies one — `IncompatibilityConstraint::enforcing()` adds approval obligations for the capabilities the meet produces. Measured field by field:

```
a            = restrictive                    obligations {}
a ∧ (a ∨ b)                                   obligations {RunBash, GitPush, CreatePr}
```

capabilities, paths, budget, commands, time and isolation all agree; only obligations differ. Those are the uninhabitable-state constraint doing its job, so *"make absorption hold"* would mean weakening it.

So the type is `j(a ∧ b)` for a closure `j` — a **nucleus** on a lattice, not a lattice meet — while implementing `Lattice`, `BoundedLattice` and `DistributiveLattice`, three traits whose laws it does not satisfy. This repo already has the vocabulary for what it actually is: `frame::Nucleus`, and the `ConstraintNucleus` that `scripts/law-mechanisms-manifest.txt` records as declared-dead, with production using *"hardcoded ifs"* instead.

A test records the gap so it stops being invisible, and says so if it ever closes. Reconciling the traits with the structure is a design change and is deliberately not in this PR.

## Enforcement was sound throughout

`leq` is unchanged and was never affected — it compares only the policy fields. The one production enforcement site, `certificate.rs`'s `effective_permissions.leq(prev_permissions)`, was correct before and after.

## Verification

`cargo test -p portcullis` — 1052 lib tests plus every integration suite, all green. `cargo clippy -p portcullis --all-targets --all-features -- -D warnings` clean. The scorecard is unchanged at `life 0.00%`; this PR fixes a defect the `alg` family surfaced but does not yet declare the laws, because three of them still do not hold.

### DIFF
--- a/.scorecard-ratchet.toml
+++ b/.scorecard-ratchet.toml
@@ -136,13 +136,21 @@ population_floor = 278
 # looked and it was fine". They are reported as undeclared and the population is
 # the 80 crates clippy can see.
 #
+# 2026-09-11 (second pass): nine more crates cleared and annotated, 13 -> 22.
+# Each was ONE arithmetic site -- a Vec::with_capacity hint or a counter -- where
+# saturating_add/mul is exactly equivalent: a capacity hint that saturates
+# allocates rather than wraps, and the loop fills the real size. 24 more crates
+# sit within 8 sites of clean; clearing all of them would reach 46 of 80 (57.5%).
+# The remainder need judgement about a contract rather than an arithmetic
+# rewrite, so they are not batched in here.
+#
 # 2026-09-11: seeded after annotating the 13 crates that were already clean, and
 # probing one of them red -- `v[0] + 1` in nucleus-pca fires indexing_slicing and
 # arithmetic_side_effects, and the crate is green again when it is removed. The
 # 67 remaining crates need their sites fixed before they can declare; the two
 # tractable lints there are unwrap_used (87) and panic (10), and the 1528
 # arithmetic and indexing sites are a program, not a gate.
-floor_bp = 1728
+floor_bp = 2750
 population_floor = 81
 
 [family.life]
--- a/.scorecard-ratchet.toml
+++ b/.scorecard-ratchet.toml
@@ -150,7 +150,7 @@ population_floor = 278
 # 67 remaining crates need their sites fixed before they can declare; the two
 # tractable lints there are unwrap_used (87) and panic (10), and the 1528
 # arithmetic and indexing sites are a program, not a gate.
-floor_bp = 2750
+floor_bp = 2839
 population_floor = 81
 
 [family.life]
--- a/crates/ctf-mcp/src/main.rs
+++ b/crates/ctf-mcp/src/main.rs
@@ -9,6 +9,23 @@
 //!   - submit_attack: Submit a tool-call sequence against a level
 //!   - run_challenge: Run attacks across multiple levels in one call
 
+// ADR 0007 totality: a function whose signature says it returns is lying if it
+// panics. Denied for the shipped build only — `assert!` IS a panic, so denying
+// inside `#[cfg(test)]` would forbid the thing tests are made of. This is the
+// same line `is_production_path` draws when it strips the test region.
+#![cfg_attr(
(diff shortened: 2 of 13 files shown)


brandon-coproduct:
Both reds here are **one cause**, and it is not what the gate says it is.

## The arithmetic

`.scorecard-ratchet.toml` pins `[family.tot] floor_bp = 2750`. **27.50% is exactly 22/80.** The tree measures **27.16%, which is exactly 22/81.**

Reproduced on a clean detached checkout of this branch — same numbers as CI, so it is not an environment difference. With the floor lowered to the measured value the card prints its operands:

```
  family  unit                              population  discharged         %  undeclared
  tot     production crate                          81          22    27.16%           5
```

**discharged is 22 on both sides. Nothing stopped being discharged.** The population went 80 → 81: one crate joined the measurable surface and does not declare totality, so the ratio fell at a constant numerator.

The gate's message is *"an obligation the tree declares stopped being discharged"*. That sends a reader looking for a removed `#![deny]`, which is not there.

## It is not confined to this PR

| PR | floor_bp | = | measured | = |
|---|---|---|---|---|
| #2835 | 1625 | **13/80** | 1604 | **13/81** |
| #2838 | 2750 | **22/80** | 2716 | **22/81** |
| #2840 | 2750 | 22/80 | 2716 | 22/81 |

Two different numerators, both landing exactly on *d*/81 against floors that are exactly *d*/80. One crate, three PRs, six red checks.

## The structural part

`population_floor` is a **minimum**, so the population may grow freely — and growth is precisely what breaks a pinned ratio. The file's header names the two traps it was built for (*"a floor on the ratio rewards deleting obligations"*, *"a pin on the debt misses a ratio falling at constant debt"*) and this is the third: **the ratio falling because the surface grew.** The fix is for `Finding::Fell` to compare `discharged` as well and say which operand moved — at present it can only name one of two causes, and it names the wrong one here.

Immediate unblock is either re-pinning `floor_bp` to the measured 2716 with a dated note, or declaring totality on the 81st crate to get 23/81 = 2839. Which of those is right is yours; the gate is correct to refuse, it just misattributes.

## And the third red

`Gates must fail on their own subject` is a *consequence* of the same thing: `probe_xtask` measured `xtask scorecard` perturbed and restored, and read the red restore as *"the perturbation left something behind"* — four times, about a gate that was red on arrival. #2870 measures the baseline first, so an already-red gate says so and stops.

Issue #747, opened 2026-04-01
AUDIT: Receipt chain hash uses format!("{:?}") for verdict — Debug repr is not stable


brandon-coproduct:
## Summary

\`VerdictReceipt::compute_hash()\` in \`receipt_chain.rs\` line 71:

```rust
hasher.update(format!("{:?}", self.verdict).as_bytes());
```

The hash uses Rust's \`Debug\` formatting of the \`FlowVerdict\` enum. The \`Debug\` representation is not a stable ABI — it can change between Rust compiler versions, or when the enum is modified (e.g., adding a field to a variant). This means:

1. Receipt hashes computed by different compiler versions may differ for the same logical verdict
2. A chain exported from one binary version cannot be verified by another
3. Adding a field to \`FlowDenyReason\` silently breaks all existing chains

## Impact

**MEDIUM** — Receipt chain interoperability and long-term archival integrity. Chains are supposed to be tamper-evident logs that can be verified years later. Using \`Debug\` formatting for hash computation makes them fragile.

## Files

- \`/crates/portcullis/src/receipt_chain.rs\` — line 71

## Remediation

1. Use a deterministic, versioned encoding for the verdict (e.g., a discriminant byte + fixed-length fields)
2. Or: use serde JSON serialization with \`#[serde(tag = ...)]\` for a stable representation
3. Add a golden-value test that pins the hash output for known verdicts

PR #2840, opened 2026-09-11
A decision may only be redeemed under the permissions it was taken against

brandon-coproduct:
The first step of LIFE's validity work, built after establishing what the redeem path actually does.

## The gap

The affine discipline on `DecisionToken` proves it **cannot be used twice**. It proved nothing about whether it is still **true**, and the redeem-side check was narrower than it looks:

```rust
fn require_decision_for(decision_op: Operation, performing: Operation) {
    if decision_op == performing { return Ok(()) }
    ...
}
```

Two `Operation`s compared. The kernel is not consulted at all — and `DecisionToken` is only ever *minted* in portcullis; there is no kernel-side redeem anywhere. **A token decided under one policy was redeemable under any other.**

## The fix

The token carries the checksum of the effective permissions it was decided against; the executor — which already holds the policy it will act under — refuses a token whose checksum is not that one. Six redeem sites, all wired.

**Normalized on both sides.** The kernel decides against `self.effective` as written; an executor enforces `policy.clone().normalize()`, and `normalize` is not the identity — it applies the uninhabitable-state constraint, which *adds* approval obligations. Comparing raw checksums refused every token for a policy that normalization changes, which is how two existing tests caught the mistake.

## It required making the checksum mean something first

`PermissionLattice::checksum` serialized the whole struct — `id`, `description`, `derived_from` included — so **two values that compare equal hashed differently**. That is the `Hash`/`Eq` coherence law, broken, and it had a consequence: a `meet` producing the same policy under a new label produced a different digest, so the audit chain recorded `pre_permissions_hash != post_permissions_hash` — *a permission change that had not happened*.

The non-serde variant hashed `format!("{:?}", self)`. Derived `Debug` is not a stability contract — the defect #747 records for the receipt chain, here on the permission checksum itself. Both variants now hash the same eight policy fields, tagged and length-prefixed.

**`effective_permissions_detect_tampering` failed on that change, and it was right.** A sealed structure seals everything it carries, including the audit label a reviewer reads. So `EffectivePermissions` got its own seal over the whole value, provenance included. Two questions, two digests — *what is permitted* and *is this the exact value I sealed*. Collapsing them is what made the checksum incoherent with equality in the first place.

## What this bounds, exactly

**Not elapsed time, but whether the thing the decision depended on is the thing being executed under.** A state fingerprint is a *tighter* bound than a TTL, not a weaker one: it expires the instant the state changes rather than on a timer someone guessed, and it imports no clock and no host/guest skew across the vsock boundary.

It does **not** yet catch a budget spent or an approval consumed between decision and effect — those are kernel state the executor does not hold — and it does not re-read the kernel, so a policy attenuated after the executor was built is unseen. Both are the next step, and both are cheap now the fingerprint is on the token.

## The scorecard

`life` goes **0 → 1 of 7**, and the measured zero did exactly what it was admitted to do. It did not sit quietly while the work happened:

```
FAIL: 1 scorecard finding(s).
  life is at 14.28% but the floor is pinned at 0.00%. Raise floor_bp to the
  measured value: a floor with slack under it has already stopped gating.
```

That is the whole defence of admitting `floor_bp = 0`, driven rather than argued.

**One bug in that family, found by it failing to see its own subject.** `strip_visibility` did not exist — only `pub ` was stripped — so every `pub(crate)` field was invisible, including `pub(crate) permissions` on `DecisionToken`. The gate reported the work as not done.

## Verification

Probed both ways on the real types, not on two strings:

| case | result |
|---|---|
| token minted under policy A, executor running policy B | refused, message naming both digests |
| token minted under policy A, executor running policy A | accepted — so the refusal is not passing by refusing everything |

`cargo test -p portcullis` — 1054 lib tests and 22 suites green. `cargo test -p nucleus --lib` — 81 green. `cargo clippy -p portcullis -p nucleus -p xtask --all-targets --all-features -- -D warnings` clean. 224 xtask tests, including three new ones pinning that a restricted-visibility field is still a field, that a state fingerprint bounds validity, and that identifying a right (`session_id`, `sequence`, `issued_at`) is not bounding it.

### DIFF
--- a/.cargo/mutants.toml
+++ b/.cargo/mutants.toml
@@ -1,5 +1,11 @@
 # cargo-mutants configuration.
 #
+# In .cargo/, not the repo root. `cargo mutants --help` says so plainly —
+# "Read configuration from this file instead of .cargo/mutants.toml" — and a
+# root-level mutants.toml is read by nothing. The first version of this file was
+# at the root, CI reported the same two survivors unchanged, and `--list` locally
+# confirmed the exclusion was doing nothing at all.
+#
 # WHY THIS FILE EXISTS
 #
 # `Mutation Testing` runs `cargo mutants --package portcullis --all-features`.
--- a/.scorecard-ratchet.toml
+++ b/.scorecard-ratchet.toml
@@ -190,10 +190,29 @@ population_floor = 81
 # sentence rather than a value someone left out, and the parser refuses the key
 # once floor_bp rises above zero, so the acknowledgement cannot go stale.
 #
-# 2026-09-11: seeded at the measurement. Giving a right an expiry is a behaviour
-# change on a security boundary -- a clock at the mint site, a check at the
-# consume site, and a decision about what expiry means for an operation already
-# in flight. This gate states the debt exactly; it does not pay it.
-floor_bp = 0
+# 2026-09-11 (second pass): DecisionToken carries the checksum of the effective
+# permissions it was decided against, and the executor refuses a token whose
+# checksum is not the one it is about to run under. 0 -> 1 of 7.
+#
+# The measured zero did exactly what it was admitted to do. It did not sit
+# quietly at zero while the work happened: the slack check turned the first
+# discharge into a red -- "life is at 14.28% but the floor is pinned at 0.00%"
+# -- and the pin could not be carried forward. That is the whole defence of
+# admitting floor_bp = 0, and it was driven rather than argued.
+#
+# A STATE FINGERPRINT IS A TIGHTER BOUND THAN A CLOCK, not a weaker one. A TTL
+# asks "how long has it been?"; the invariant a one-shot right needs is "has
+# anything it depended on changed?". Those diverge both ways -- a thirty-second
+# token is stale at one millisecond if the policy was amended, and sound an hour
+# later if nothing moved -- and a TTL also imports a trusted clock and host/guest
+# skew across the vsock boundary. So the family counts either form, and the
+# vocabulary says which is which.
+#
+# Not yet bounded: Authority, CheckProof, DischargedBundle, PodRuntime,
+# ServeToken, VerifyToken. The last two cross a process boundary and are
+# serialized, so they need the fingerprint checked on arrival rather than a
+# borrow; the first four are in-process and could take a lifetime instead, which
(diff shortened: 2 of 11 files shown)


brandon-coproduct:
`Manifest Guards` is red here, and it isn't this PR.

```
FAIL: 1 scorecard finding(s).
  tot fell to 27.16% against a floor of 27.50%: an obligation the tree declares stopped being discharged.
```

**Same failure, same two numbers, on a branch that isn't flagged red.** I checked out `adr/bound-remaining-rights` and ran `cargo xtask scorecard` directly:

```
CLEAN_BRANCH_SCORECARD_RC=1
FAIL: 1 scorecard finding(s).
  tot fell to 27.16% against a floor of 27.50%
```

So the floor is unsatisfiable on the shared base rather than regressed by any one of these PRs. #2838 and #2835 show the same red.

**It cannot red `main` yet**, which is why this is a note and not an alarm: `scorecard` is not a subcommand on `main` (`error: unrecognized subcommand 'scorecard'`), and `main`'s `manifest-guards.yml` does not call it. I checked all six queued PRs — none carries the gate. So it lands with the first of this family to merge, and reds on arrival unless the floor moves with it.

The branches are only **2 commits behind main**, so staleness isn't the explanation either — I tested that first.

I haven't tried to work out which change lowered `tot`, because that is inside work you're actively doing and the answer is probably obvious to you and not to me. Flagging the shape: **a ratchet floor raised on one branch fails every sibling that lacks the numerator**, and here every sibling fails including the one that carries the raise.

brandon-coproduct:
Same cause as #2838 — full arithmetic there: https://github.com/coproduct-opensource/nucleus/pull/2838#issuecomment-5643371289

Short version: `[family.tot] floor_bp` on this branch is exactly *d*/**80**, and the tree measures exactly *d*/**81**. `discharged` is unchanged on both sides — the population grew by one crate, so the ratio fell at a constant numerator. The gate says *"an obligation the tree declares stopped being discharged"*, which sends you looking for a removed `#![deny]` that is not there.

Here it is `2750 = 22/80` pinned against `2716 = 22/81` measured — inherited from #2838, this branch adds no `tot` work.

`Gates must fail on their own subject` is a consequence of the same red, misreported as a restore failure; #2870 fixes that message.

PR #2843, opened 2026-09-11
A per-pod capability is bound to its pod, not merely documented as

brandon-coproduct:
`BrokerCapability::mint` has always carried this:

> *Per-pod and never reused: two pods sharing one would let either sign for the other, and the listener's identity binding — one socket per VM — would stop meaning anything.*

**Nothing enforced it.** `ServeToken` and `VerifyToken` held a bare `String`, and `into_served` / `into_verifier` handed it to whoever called. Stating an invariant in prose beside a method that cannot check it is the same shape as `DecisionToken::operation` two commits ago: a consuming accessor that reads a right without validating it.

Both tokens now carry the pod they were minted for, `mint` takes it, and the consuming accessors require it and return `Result<_, WrongPod>`. The broker listener refuses to start on a mismatch rather than authenticating another pod's proxy against this pod's broker.

## Not a `life` change — the scorecard stays at 1 of 7

A pod-bound token never expires *for its own pod*. It is a **scope** bound — the same kind `DecisionToken`'s operation already was — not a validity bound. Adding `pod` to the family's vocabulary would move the number without bounding anything, which is the self-serving widening the family's own doc warns against.

## A guard's needle drifted again

`mint_appears_exactly_once_on_the_spawn_path` greps main.rs for `BrokerCapability::mint()` — with empty parens. Giving the constructor an argument would have left it matching **zero**, and a guard that counts zero either fails loudly or gets "fixed" by deletion. It now greps `mint(`.

That is the **third time in this campaign** a structural guard's needle has drifted from the code it guards.

`two_pods_do_not_share_a_capability` asserted two mints produce different secrets; it still does, plus two new tests covering the refusal and — so the refusal is not passing by refusing everything — the acceptance.

## Verification, stated honestly

The token mechanism is platform-independent and covered by four tests that run here: **482 nucleus-node tests green, clippy clean.**

The three call sites are inside `#[cfg(target_os = "linux")]` and **cannot be compiled on this machine**. Removing the cfg fails on sibling modules that are also gated; cross-checking via `--target aarch64-unknown-linux-musl` fails in `ring`'s build script for want of a C cross-compiler. Their correctness is three type-checked lines resting on CI's compile, and that is the weakest link in this change — flagging it rather than implying the whole thing was verified locally.


brandon-coproduct:
All six reds here are **one cause, and it is not in your diff**: a 134-second race between this PR's CI and its parent merging.

```
BASE_REF: adr/decision-lifetime
fatal: invalid object name 'origin/adr/decision-lifetime'.
fatal: ambiguous argument 'origin/adr/decision-lifetime...HEAD': unknown revision
##[error]Process completed with exit code 128.
```

| | |
|---|---|
| these runs created | **07:18:28Z** |
| #2842 merged, deleting `adr/decision-lifetime` | **07:20:42Z** |

The jobs had already captured `BASE_REF` and resolved `origin/$BASE_REF` after the branch was gone, so every job that diffs against the base died with git 128 — `Detect Changed Crates`, `musl type-check`, `A2A example server`, `Constitutional Gate`, the fmt shadow. GitHub has since retargeted this PR to `adr/decision-token`, which still exists.

I have re-run the failed jobs. Per the standing distinction: this is damage **about the run**, not **in the branch**, so a rerun is the right tool rather than `update-branch` — 3 runs instead of ~30.

**What remains yours:** the PR is `DIRTY` against its new base. That needs a rebase onto `adr/decision-token`, and no rerun will clear it.

The general shape, recorded as gatehouse F-147: **a stacked PR's in-flight CI dies when its parent merges**, because the base branch is deleted out from under jobs that already read its name. GitHub's auto-retarget fixes the PR but not the runs, and the resulting six red checks look like six defects.

brandon-coproduct:
**Correcting my own advice above: the rerun cannot fix this, and I should not have said it would.**

I re-ran the failed jobs. They still fail, with the same line:

```
BASE_REF: adr/decision-lifetime
```

— the branch that no longer exists. **A rerun replays the original event payload**, so it faithfully reproduces the stale base ref rather than picking up GitHub's retarget to `adr/decision-token`. Only a *new trigger* — a push to the branch, or the rebase this PR needs anyway — produces a run that sees the new base.

That is a third category the standing distinction did not have. F-117 separated damage **in the branch** (needs a new commit) from damage **about the run** (a rerun suffices); this is damage **in the event**, and a rerun is precisely the tool that cannot repair it, because reproducing the event is what it does.

So the diagnosis stands unchanged — six reds, one 134-second race, no defect in your diff — but the remedy is only the rebase onto `adr/decision-token`, which this PR needs regardless of being `DIRTY`. The three reruns I spent were wasted; nothing else was affected.

Recorded as gatehouse F-147 with the correction.

PR #2846, opened 2026-09-11
fix(test): registry tests shared a temp dir, so one pulled another's manifest

brandon-coproduct:
`Tests` went red on an unrelated PR (#2825, which touches nothing within three crates of this):

```
assertion `left == right` failed
  left: Some("{\"data\":\"value\"}")
 right: Some("{\"schema_version\":\"1\",\"layers\":[]}")
```

`push_pull_roundtrip` pushed the manifest on the **right** and pulled back the one on the **left** — which belongs to `pull_by_digest`, forty lines below it. **Two tests were using the same directory.**

## Why

The helper keyed the directory on `SystemTime::now().as_nanos()` alone. The name says nanoseconds; the clock does not supply them. Measured, 200,000 consecutive calls:

| | |
|---|---|
| identical consecutive values | **196,175 / 200,000** (98.1%) |
| smallest non-zero delta | **1000 ns** |

It is a microsecond counter wearing a nanosecond type. Two tests starting in the same microsecond get the same path — and CI runs **nextest, one process per test**, on 8 cores.

## The fix

Three components, because **no one of them is sufficient**:

- **PID** — separates nextest's one-process-per-test, where an in-process counter cannot help.
- **atomic counter** — separates tests sharing a process under plain `cargo test`, where the PID cannot help.
- **timestamp** — separates runs.

Uniqueness by construction, rather than by hoping two threads never land on the same tick.

## Verification (A-19)

Driven RED on the real defect. 64 calls to each helper:

```
OLD tempdir: 17/64 distinct  -> test would FAIL
NEW tempdir: 64/64 distinct  -> test would PASS
```

The added `tempdir_is_unique_within_a_process` is the falsifier, and it did not exist before. 25 registry tests pass; clippy and fmt RC=0.

## Worth naming

This is a **flake**, which is the failure mode that costs most per unit of defect: it fails somebody else's PR, at random, for a reason their change cannot explain — and the cheapest response is to re-run rather than read. Nine tests in this file use the helper, so any pair of them could have collided.

### DIFF
--- a/crates/portcullis-core/src/registry.rs
+++ b/crates/portcullis-core/src/registry.rs
@@ -575,9 +575,32 @@ mod tests {
 
     // ── helpers ────────────────────────────────────────────────────
 
+    /// A directory no other test shares.
+    ///
+    /// This used to key on `SystemTime::now().as_nanos()` alone, which is not unique: the
+    /// name says nanoseconds but the clock does not supply them. Measured 2026-09-11 on
+    /// macOS, 196,175 of 200,000 consecutive calls returned an IDENTICAL value and the
+    /// smallest non-zero delta was 1000 ns — it is a microsecond counter wearing a
+    /// nanosecond type. Two tests starting in the same microsecond therefore got the same
+    /// directory, and one pulled the other's manifest:
+    ///
+    /// ```text
+    /// assertion `left == right` failed
+    ///   left: Some("{\"data\":\"value\"}")            <- pull_by_digest's manifest
+    ///  right: Some("{\"schema_version\":\"1\",\"layers\":[]}")
+    /// ```
+    ///
+    /// Three components, because no one of them is sufficient: the PID separates nextest's
+    /// one-process-per-test (where an in-process counter cannot help), the counter separates
+    /// tests sharing a process under plain `cargo test` (where the PID cannot help), and the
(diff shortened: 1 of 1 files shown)


brandon-coproduct:
Closing as a duplicate of #2847, which lands the same fix and does it **better**: pid + counter with **no clock at all**, where mine kept a timestamp as a third component. The clock is the thing that failed, and there is no reason to keep a share of it — `remove_dir_all` already makes path reuse across runs harmless, and pid differs between concurrent runs on one machine, so pid+counter is unique by construction.

#2847's analysis is also sharper than mine on the mechanism: it is not merely that two tests *share* a directory, it is that the second one's `remove_dir_all` **deletes the first one's registry underneath it, mid-test** — and that re-running reproduced it, making it a race under load rather than a flake that goes away.

Two things from here worth carrying over, left as a comment on #2847 rather than lost:

- the clock measurement (196,175 of 200,000 consecutive `as_nanos()` calls identical; smallest non-zero delta 1000 ns), which is the evidence that the old key was never unique rather than unluckily unique;
- `tempdir_is_unique_within_a_process`, the falsifier — #2847 has no test, and 64 calls to the old helper yield 17 distinct paths, so the property is cheap to pin and was demonstrably falsifiable.

Issue #91, opened 2026-02-24
Flaky test: test_sign_csr_with_key fails with NomError(Eof)


brandon-coproduct:
## Description

`ca::self_signed::tests::test_sign_csr_with_key` intermittently fails with:

```
called `Result::unwrap()` on an `Err` value: CsrGeneration("failed to parse CSR: Parsing Error: NomError(Eof)")
```

Location: `crates/nucleus-identity/src/ca/self_signed.rs:690`

## Impact

Blocks `git push` via pre-push hook since the full test suite fails. Currently requires `--no-verify` to push unrelated changes.

## Root Cause Hypothesis

The CSR generation or serialization produces a truncated DER blob in certain conditions. The `x509-parser` nom-based parser then fails at Eof because the DER encoding is incomplete. This may be a race condition in key generation or a subtle incompatibility between `rcgen` CSR generation and `x509-parser` deserialization.

## Suggested Fix

1. Add explicit length validation after CSR generation before parsing
2. Check if `rcgen`/`x509-parser` version combination has known incompatibilities
3. Consider wrapping the test in a retry or adding better error context

PR #2847, opened 2026-09-11
fix(registry): nine tests shared a directory named by the clock

brandon-coproduct:
`tempdir()` in `portcullis-core::registry`'s test module built its path from `SystemTime::now().as_nanos()` and then `remove_dir_all`-ed it. **Nine tests in that module call it.** Two landing in the same nanosecond bucket get the same path, and the second one's `remove_dir_all` deletes the first one's registry underneath it, mid-test.

## How it was found

`push_pull_roundtrip` failed on #2825 — a PR whose entire diff is a workflow timeout — as **1 failure in 7444**.

The failure was invisible from the job that reported it. `Tests` contains only a `Report` step relaying `STEP: failure` from `Workspace build + tests`, a job whose own conclusion is `success` because the step carries `continue-on-error`. So the GitHub API never lists it as failed, and the test name was three jobs away from the red mark.

**I re-ran it first**, on the theory it was the known `NomError(Eof)` flake (#91), and said at the time that was a guess rather than a diagnosis. It reproduced — which is what a clock-keyed name does under load. It is a race, not noise.

## The fix

Process id **and** a counter, because the two runners collide differently:

- `cargo nextest` gives each test its own process, so the pid alone separates them
- plain `cargo test` runs them as threads in one process, so the pid is shared and only the counter does

Neither reads a clock, so neither can collide.

## Verification

`cargo test -p portcullis-core --all-features --lib registry::` — 54 passed, run three times. Clippy clean.

This is on `main` rather than in my PR stack because it affects every PR's `Tests` job, not just mine.

### DIFF
--- a/crates/portcullis-core/src/registry.rs
+++ b/crates/portcullis-core/src/registry.rs
@@ -575,13 +575,30 @@ mod tests {
 
     // ── helpers ────────────────────────────────────────────────────
 
+    /// A directory no other test can name.
+    ///
+    /// This used to be `temp_dir().join(format!("nucleus-registry-test-{nanos}"))`
+    /// followed by `remove_dir_all`. Nine tests in this module share it, and two
+    /// landing in the same nanosecond bucket get the SAME PATH — at which point
+    /// the second one's `remove_dir_all` deletes the first one's registry
+    /// underneath it, mid-test.
+    ///
+    /// That is not hypothetical. `push_pull_roundtrip` failed in CI twice on one
+    /// PR whose diff was a workflow timeout, 1 failure in 7444, and re-running
+    /// reproduced it: a clock-keyed name is a race that fires under load, not a
+    /// flake that goes away.
+    ///
+    /// Process id AND a counter, because the two runners collide differently.
+    /// `cargo nextest` gives each test its own process, so the pid separates
+    /// them; plain `cargo test` runs them as threads in one process, so the pid
+    /// is shared and only the counter does. Neither reads a clock.
     fn tempdir() -> PathBuf {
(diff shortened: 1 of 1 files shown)


brandon-coproduct:
Closed #2846 in favour of this — your fix is the better one. Dropping the clock entirely is right; I had kept a timestamp as a third component, and there is no reason to keep a share of the thing that failed.

Two pieces of evidence from that branch, offered here rather than lost.

**1. The old key was never unique, not unluckily unique.** Measured over 200,000 consecutive calls:

| | |
|---|---|
| identical consecutive values | **196,175 / 200,000** (98.1%) |
| smallest non-zero delta | **1000 ns** |

`as_nanos()` is a microsecond counter wearing a nanosecond type. That turns "two tests landing in the same nanosecond bucket" from an unlucky coincidence into the ordinary case — which matches your observation that re-running reproduced it.

**2. A falsifier, which this PR does not have.** 64 calls to the *old* helper yield **17** distinct paths; to the new one, **64**. So the property is cheap to pin and was demonstrably falsifiable:

```rust
/// The property the old helper lacked: two calls never collide, even when the
/// clock does not move between them.
#[test]
fn tempdir_is_unique_within_a_process() {
    let n = 64;
    let dirs: std::collections::BTreeSet<PathBuf> = (0..n).map(|_| tempdir()).collect();
    assert_eq!(dirs.len(), n, "tempdir() handed out a duplicate path");
}
```

It passes against your implementation and fails against the one being replaced. Take it or leave it — but a fix whose regression is this cheap to pin seems worth pinning, and it is the one thing #2846 had that this does not.

Issue #1216, opened 2026-04-05
feat: enforcement completeness — Clippy disallowed_types gates all I/O in hook and MCP surfaces


brandon-coproduct:
## Problem

`portcullis-effects` (#1191) provides `PolicyEnforced<RealEffects>` as the policy-checked I/O surface. But the enforcement is voluntary: `nucleus-claude-hook` and `nucleus-mcp` currently call `std::fs`, `std::process::Command`, and `reqwest` directly in dozens of places. The effect layer exists but is not structurally required.

The flagship safety claim — "no external side effect unless mediated by Nucleus and authorized by policy" — is false until every I/O call site in the agent surface goes through `PolicyEnforced`.

### Clippy `disallowed_types` configuration

Add to `.cargo/config.toml` (or per-crate `Cargo.toml` under `[lints.clippy]`):

```toml
# nucleus-claude-hook and nucleus-mcp: all I/O must go through portcullis-effects
[lints.clippy]
disallowed-types = [
    { path = "std::fs::File", reason = "use portcullis_effects::FileEffect" },
    { path = "std::fs::OpenOptions", reason = "use portcullis_effects::FileEffect" },
    { path = "std::process::Command", reason = "use portcullis_effects::ShellEffect" },
    { path = "tokio::process::Command", reason = "use portcullis_effects::ShellEffect" },
    { path = "reqwest::Client", reason = "use portcullis_effects::WebEffect" },
    { path = "reqwest::blocking::Client", reason = "use portcullis_effects::WebEffect" },
]
```

### Migration of existing direct I/O

Audit all `std::fs::*` and `std::process::*` calls in:
- `crates/nucleus-claude-hook/src/*.rs` (~40 call sites across main.rs, build.rs, config.rs, etc.)
- `crates/nucleus-mcp/src/main.rs` (~5 call sites)

For each call site:
1. Determine which `CapabilityLattice` field governs it
2. Route through an injected `impl FileEffect` / `impl ShellEffect`
3. Replace the call with the effect method

### Enforcement in CI

The Clippy job already runs with `-D warnings`. Once `disallowed_types` is configured, any new direct I/O import in the affected crates is a CI failure — enforcement is automatic and permanent.

### Exemptions (documented, not silenced)

[...]

PR #2850, opened 2026-09-11
suppress: the exemption channel of every other family

brandon-coproduct:
```
  scorecard | suppress 3.73%

  family    unit                              population  discharged         %  undeclared
  bound     witness-accepting parameter site         172         171    99.41%          32
  alg       (lattice type, law) obligation           278         110    39.56%           8
  tot       production crate                          80          22    27.50%           5
  life      affine right                               7           1    14.28%           0
  typed     runtime flow kind                         23           2     8.69%           0
  suppress  lint suppression                         134           5     3.73%          53
```

## Why this is not one gate among six

`bound`, `alg`, `tot`, `life` and `typed` each report how much of what the tree declares is covered by *a mechanism that can fail*. Every one of those mechanisms can be switched off, silently and permanently, by one line:

```rust
#[allow(clippy::let_underscore_must_use)]
```

An `#[allow]` is forever. It outlives the reason it was written for and nothing says when that reason expired. An `#[expect]` is the same waiver with an expiry attached: when the lint stops firing there, `unfulfilled_lint_expectations` errors and the attribute has to go. **It is the only suppression that can itself fail** — which is this card's definition of *discharged*, applied to the waivers.

## The asymmetry is the case

**292 `#[allow]` across the tree, of which zero carry a `reason =`. 48 `#[expect]`, of which all 48 do.** The tree already knows the right form and 86% of its suppressions are the other one.

## Production only, and here that is not merely the convention

A suppression inside `#[cfg(test)]` cannot disable a production gate, and disabling a production gate is the defect being measured. That is why the number is 5 of 134 rather than 39 of 206 — most existing `#[expect]`s waive lints at test sites, including nine of `portcullis-effects`' ten.

## A mandate only reaches where its table does

`allow_attributes` arrives through `[workspace.lints.clippy]`, which reaches a crate only if it opts in with `[lints] workspace = true`.

**54 of 95 crates do. 41 do not — including `portcullis`, `portcullis-core` and `nucleus-flow-replay`**, the crates holding the lattice, the receipt hashing and the flow replay. Their 53 production suppressions are `undeclared`, not discharged, because the alternative is that a crate could *leave* the lint table and have its allows become compliant. Same reading `tot` takes of the crates clippy cannot compile (ADR 0007 A-2).

## How it gets paid

Two lines into the existing `[workspace.lints.clippy]` table — `allow_attributes = "deny"` and `allow_attributes_without_reason = "deny"` — and `clippy::allow_attributes` has a machine-applicable suggestion, so most of the conversion is `cargo clippy --fix`. Two real costs: an `#[allow(dead_code)]` that fires only under some feature combinations becomes an *unfulfilled expectation* under others, so those need per-site judgement under the `--all-features` matrix; and the 41 non-adopting crates need three manifest lines each before the mandate reaches them at all.

## Two bugs the tests found while writing this

`adopting_crates` keyed on `"\n[lints]"`, so a manifest whose **first** line is `[lints]` would have been scored outside a mandate it is inside. And the adapter test read real manifests through `git ls-files`, which fails under `cargo test` because the working directory is the crate root, not the repo root — so the `Report → Census` mapping is now a pure `const fn` that tests without a checkout.

## The first probe proved nothing

I downgraded an `#[expect]` inside a `#[cfg(test)]` block and no number moved, because the test region is stripped. The registered probe uses `crates/nucleus-tool-proxy/src/art12.rs`, where `#[expect(clippy::disallowed_methods, reason = "#1216 ...")]` guards the proxy's own Article 12 record. Downgrading it takes `suppress` 3.73% → 2.98%.

246 xtask tests green, clippy clean.

### DIFF
--- a/.scorecard-ratchet.toml
+++ b/.scorecard-ratchet.toml
@@ -264,3 +264,52 @@ population_floor = 7
 # Secret are confidential -- and each is a struct plus a registry row.
 floor_bp = 869
 population_floor = 23
+
+[family.suppress]
+# 5 of 134 production lint suppressions expire. The rest are forever.
+#
+# THIS FAMILY MEASURES THE EXEMPTION CHANNEL OF THE OTHER FIVE. bound, alg, tot,
+# life and typed each report how much of what the tree declares is covered by a
+# mechanism that CAN FAIL. Every one of those mechanisms can be switched off,
+# silently and permanently, by one line: #[allow(clippy::whatever)].
+#
+# An #[allow] is forever. It outlives the reason it was written for and nothing
+# says when that reason expired. An #[expect] is the same waiver with an expiry:
+# when the lint stops firing there, unfulfilled_lint_expectations errors and the
+# attribute has to go. It is the ONLY suppression that can itself fail, which is
+# this card's definition of discharged applied to the waivers.
+#
+# THE ASYMMETRY IS THE CASE. Across the tree: 292 #[allow], of which ZERO carry a
+# `reason =`. 48 #[expect], of which all 48 do. The tree already knows the right
+# form and 86% of its suppressions are the other one.
+#
+# PRODUCTION ONLY, and for this family that is not merely the convention. A
+# suppression inside #[cfg(test)] cannot disable a production gate, and disabling
+# a production gate is the defect being measured. That is why the number is 5 of
+# 134 rather than 39 of 206: most existing #[expect]s waive lints in test code,
+# including nine of portcullis-effects' ten, which waive
+# clippy::let_underscore_must_use at test sites.
+#
+# A MANDATE ONLY REACHES WHERE ITS TABLE DOES. allow_attributes and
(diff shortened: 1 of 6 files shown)


brandon-coproduct:
Closing as already included, not as rejected.

`git diff origin/adr/typed-family origin/adr/suppress-family` is empty — the two branches have identical trees — and `adr/typed-family` (#2845, this PR's own base) already carries `4c8807f03 feat(suppress): the exemption channel of every other family (#2850)`. The squash landed on the base branch rather than main, so GitHub never closed this.

The suppress family therefore reaches main through #2845, with the scorecard gate passing there:

```
OK: 6 families on the card, weakest is suppress at 3.73%.
  suppress  lint suppression   134   5   3.73%   53
```

Nothing is lost by closing this; an empty PR cannot go green, so leaving it open only blocks the stack behind it. Reopen if the separation was deliberate.

PR #2856, opened 2026-09-12
The shadow gate's timeout fires before its job dies, and a red says what it means

brandon-coproduct:
The nucleus half of gatehouse **F-100**. The gatehouse runner deadlocked on any command writing more to stderr than a pipe buffer holds — which is every cargo command there is — and hung all three shadow legs on **sixteen** open PRs for 45 minutes each.

The deadlock is fixed in gatehouse (one reader thread per stream; 9-test suite now runs in 1.21s, was hitting a 10s timeout). These are the three things on *this* side that turned a bug into an **undiagnosable** one.

### 1. The timeout could not fire

`timeout: "2700"` against the job's own `timeout-minutes: 45` is the same 2,700 seconds, so the runner's deadline could never win the race. The only possible outcome was a cancelled job — no receipt, no log, nothing saying why. Now `2100`, strictly below.

### 2. The runner had no outer deadline at all

The gate's `timeout_s` bounds the **command**. It says nothing about the runner hanging elsewhere — materializing the scope, reading a pipe, signing — which is exactly where it hung.

`timeout --signal=TERM --kill-after=30` at `GH_TIMEOUT + 120` turns that into exit 124 and a message that names it. The slack is deliberate: the runner must be allowed to hit its own deadline, kill the command and write an honest `errored` receipt before this fires.

### 3. Red meant "could not look" and the name said "agrees"

This check never fails on disagreement — only when the loop could not run at all. Named *"agrees with the GitHub check"*, sixteen reds read as sixteen disagreements when every one meant "could not look". I read them that way myself before checking.

gatehouse's verifier is scrupulously three-valued (`Verified` / `Refuted` / `CouldNotLook`); a GitHub check has two states. The name is the only place that distinction can survive, so it now carries it.

ADR 0007 A-2: *"I could not look" is never "I looked and it was fine"* — and it is not "I looked and it was wrong" either.

### DIFF
--- a/.github/actions/gatehouse/gatehouse.sh
+++ b/.github/actions/gatehouse/gatehouse.sh
@@ -77,8 +77,25 @@ jq -n --arg plan "$PLAN" --arg kid "$KID" --arg sk "$SK" --arg rd "$RUNNER_DIGES
     credential_tier: "developer", runner_digest: $rd, binding: null}' > "$W/job.json"
 
 # ── 4. The runner: materialize, run, sign ──────────────────────────────────────────────────
-set +e; "$RUNNER" "$W/job.json" "$W/receipt.json"; RC=$?; set -e
-case "$RC" in 0) VERDICT=held ;; 1) VERDICT=failed ;; *) die "the runner errored (exit $RC): the gate could not be run" ;; esac
+# An OUTER deadline on the runner itself, above the gate's own `timeout_s`.
+#
+# The gate's timeout bounds the COMMAND. It says nothing about the runner
+# hanging anywhere else — materializing the scope, reading a pipe, signing —
+# and on 2026-09-11 exactly that happened: a deadlock in the runner's log
+# reader (gatehouse F-100) hung every shadow job until GitHub cancelled it,
+# which produces no receipt, no log and no diagnosis. `timeout` turns that into
+# exit 124 and the message below.
+#
+# The slack is deliberate: the runner must be allowed to hit its own deadline,
+# kill the command and write an honest `errored` receipt before this fires.
+set +e; timeout --signal=TERM --kill-after=30 "$(( GH_TIMEOUT + 120 ))" \
+  "$RUNNER" "$W/job.json" "$W/receipt.json"; RC=$?; set -e
+case "$RC" in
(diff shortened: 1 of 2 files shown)


brandon-coproduct:
Heads-up on an interaction with #2825, which is currently at queue position 3. Two of its files are yours — `gatehouse.sh` and `gatehouse-shadow.yml` — and on both **your version is better**, but the third file it carries is a gate that reds on this PR.

## Where yours is better, so I'm not defending mine

- **The outer deadline has slack.** You use `GH_TIMEOUT + 120`; I used exactly `GH_TIMEOUT`. Mine races the runner's own deadline and can kill it at the moment it was about to write an honest `errored` receipt. Yours lets the runner report first, which is the whole point.
- **The name.** *"agrees with the GitHub check"* made sixteen red shadows read as sixteen disagreements when every one meant "could not look". I stared at that name for six loops and never questioned it.

## Where the gate objects

#2825 adds `cargo xtask gate-budget`, which checks `runs × timeout + setup < timeout-minutes × 60` from the declarations. Against this branch:

```
FAIL gatehouse-shadow.yml:41  job 2700s, gate 2100s x2 + 60s setup = 4260s
```

`compare` defaults to **true**, so the command runs a second time for the agreement line. At `timeout: 2100` that is 4260s of declared budget inside a 2700s job — so the runner's deadline still cannot fire first **on the second run**, which is the same hole this PR closes for the first.

Whichever of us merges second will see this red, so it isn't urgent; I'd rather it not be a surprise.

## Three ways out, and one of them is a real gap in both PRs

1. `timeout: 1200` — 2×1200+60 = 2460, fits.
2. `compare: false` for clippy — one run, 2100 fits comfortably.
3. **Bound the compare run too.** `gatehouse.sh:121` is still `set +e; bash -c "$GH_RUN"` with no bound. F-100's deadlock class applies to it as well: the plain run can hang exactly the same way, and nothing stops it. #2825 bounds it; this PR doesn't. That is the one thing mine has that yours lacks, and it's what makes the `runs × timeout` arithmetic true rather than optimistic.

My inclination is (3) plus (1), but the budget is yours to set.

PR #2857, opened 2026-09-12
Move the gatehouse pin to 89f7acc, and give five gates an image that can be pulled

brandon-coproduct:
The pin bump for gatehouse **F-100** — the runner deadlock that hung every shadow job for 45 minutes on sixteen PRs. Without this the fix reaches nothing: the shadow builds gatehouse at `GATEHOUSE_REF`, which was still the commit with the bug.

## The bump was not free, and the gate is why I know

`prelude/ci.writ` changed between the two refs. `ci.Gate` gained an appended `image` field, and `admissible_b` gained a `pinned_b` conjunct requiring every gate's image to be `<repository>@sha256:<64 hex>` pinning that gate's own env digest.

Ours were `sha256:e18a79fc...` — **a bare digest**. It pins exactly the right bytes and names no repository, so no kubelet can pull it: **all five gates in this plan could never have run.** That is the defect gatehouse wrote `pinned_b` for, and this repository is where it was found.

Fixed in both places the reference lives — `.gatehouse/gates/*.json` and the five `ci.Gate` literals in `.gatehouse/pipeline.writ` — now `docker.io/library/rust@sha256:e18a79fc...`, matching gatehouse's own gate definitions.

## Verified, including that the new conjunct bites

```
gatehouse-pin     ok: prelude/ci.writ at 89f7acc hashes to the digest the plan
                  imports (all three pins agree)

gate plan check   admissible: proved by `admissible`
                  root kernel-checked; 13 definitions, 5 gates,
                  nodes=5560 ctxs=1324 ty=4976 red=11359 sub=7962

reds-on-revert    putting one image back to the bare `sha256:` form fails
                  elaboration — NotConvertible { got: 22, want: 4 }
```

That last line is the one that matters: `pinned_b` is **not vacuous** on this plan. It refuses the exact five gates that were unpullable.

A pin bump that only moved the ref would have failed the plan check three minutes into a cargo build. `gatehouse-pin` catches the ref/digest half in milliseconds; the kernel caught the prelude half. Both were load-bearing here.


brandon-coproduct:
Since this PR is in `.gatehouse/gates/clippy.json` and `gatehouse-shadow.yml` already, one divergence it doesn't close and is cheapest to fix while you're there.

**The shadow lane's clippy entry omits a prerequisite that both other definitions of that gate include.**

| definition | command |
|---|---|
| `ci.yml:755-757` (the required check) | `wasm-pack build sdks/verifier-js --target web --release` **then** `cargo clippy --all-targets --all-features -- -D warnings` |
| `.gatehouse/gates/clippy.json` (after this PR) | `… wasm-pack build sdks/verifier-js --target web --release && cargo clippy --offline --locked …` |
| `gatehouse-shadow.yml` matrix `github:` | `cargo clippy --all-targets --all-features -- -D warnings` |

The two authoritative definitions agree with each other and the shadow entry agrees with neither — it carries only the second half. `github:` is a single shell line and GitHub's job is two steps, so the prerequisite had nowhere to go.

**It matters because that step is load-bearing, not hygiene.** `crates/nucleus-verifier-service/src/routes.rs:55` does `include_str!("../../../sdks/verifier-js/pkg/nucleus_verifier_wasm.js")`, and `git ls-files sdks/verifier-js/pkg` returns **0** — the artifact is untracked, so it cannot be materialized from any commit. Measured on a clean worktree of `main`:

```
error: couldn't read .../sdks/verifier-js/pkg/nucleus_verifier_wasm.js: No such file or directory
error: could not compile `nucleus-verifier-service` (lib) due to 2 previous errors
CLIPPY_RC=101 ELAPSED_S=58
```

So the shadow's clippy command cannot succeed on a cold checkout, on either side of the comparison it is supposed to make. The scope differs too: the spec includes `scripts/**` and `sdks/verifier-js/**`, the shadow entry includes neither — so even with the prefix restored, the sandbox has nothing to build from.

**Scoped honestly:** this is *not* why the shadow gates hung. F-100's reader-thread deadlock is, and I had explicitly noted that a compile failure surfaces in seconds while the observed behaviour was 44 minutes of silence. This is a second, quieter defect sitting behind that one — it would have surfaced the moment the deadlock was fixed, which is roughly now.

Recorded as gatehouse F-102. No action needed from me; flagging it because you're in exactly these files.

PR #2863, opened 2026-09-12
The node signs a pod receipt, with a key no guest ever holds

brandon-coproduct:
M2 step 3. `pod_receipt::Receipt` carried `v1_content_hash` — a bare SHA-256 anyone can recompute — and **no signature, and no verifier**. It said what happened and nothing said who observed it.

## Why the host signs, which is a departure

`MediationReceipt` mints a per-pod key and serves it **into the guest** over vsock. That's right for attesting decisions the in-guest mediator made. It's wrong for a receipt about what a pod produced: **a workload holding the key can sign any verdict it likes.**

SLSA Build L3's defining property is that signing keys are isolated from user-controlled build steps. `pod_authority`'s root key is persisted host-side and only *public* halves ever enter a guest — it already has that shape. This uses it.

What the signature establishes is narrow, and stated as such: *the node, holding this key, observed these bytes.* It says nothing about whether the workload told the truth in them.

## Not a signing oracle

`sign_pod_receipt` domain-tags internally rather than exposing `sign(&[u8])`. The root key also mints `LatticeCertificate`s; a general accessor would let anything holding a `&PodAuthority` have the node sign bytes of its choosing, and a receipt preimage that happened to parse as a certificate body would yield a signature valid as **both**. A test pins that a signature over the *untagged* preimage does not verify.

`verify_pod_receipt` is a **free function**, not a method. A relying party has the public key and the bytes, and must not need a `PodAuthority` — which owns the *private* key — to check a receipt. If verification required the signer, only the signer could verify.

## The preimage

Exhaustively destructured, so a new field on `Receipt` is an `E0027` until someone says whether it's committed to. Tagged and length-prefixed, no `Debug` — the discipline `check-preimage-dylint.sh` gates. `cost_usd` enters as `to_bits`, because a float's decimal rendering is a formatting decision and this is a digest.

`signature` and `signer_pubkey` bind to `_` deliberately: a signature cannot cover itself, and a test pins that setting them doesn't move the preimage.

## Seven properties

Every field moves the preimage · the two signature fields don't · framing is injective across a field boundary · another key's signature doesn't verify · an untagged signature doesn't verify · malformed hex is `false`, never a panic.

## An existing test caught the wire drift

`the_http_body_names_the_same_fields_the_proto_does` went red the moment the struct grew — the test doing its job. The HTTP body and the gRPC proto must name one vocabulary; both now carry the fields.

Moving the `ExecutionReceipt` mapping out of `main.rs` into `pod_receipt.rs` as a `From` impl is what made room under the line ceiling, and is where it belonged: the parity test compares two lists that are now both visible from one file. **main.rs: 4007 (exactly at ceiling) → 3993.**

### DIFF
--- a/crates/nucleus-node/proto/nucleus_node.proto
+++ b/crates/nucleus-node/proto/nucleus_node.proto
@@ -124,6 +124,20 @@ message ExecutionReceipt {
   uint64 cache_read_tokens = 14;
   // Estimated cost in USD (IEEE 754 double)
   double cost_usd = 15;
+
+  // === v1.2 The node's signature ===
+  //
+  // Hex Ed25519 over the receipt's tagged, length-prefixed preimage (every
+  // field above; not these two, since a signature cannot cover itself). The
+  // NODE signs — the key is the one in pod_authority and never enters a guest,
+  // which is what makes the signature worth more than the pod's own word.
+  //
+  // Empty when the node could not sign. A receipt that looks signed and is not
+  // would be worse than an unsigned one.
+  string signature = 16;
+  // The root public key that made it, hex, so a relying party can verify
+  // without asking the signer.
+  string signer_pubkey = 17;
 }
 
 message GetReceiptRequest {
--- a/crates/nucleus-node/src/main.rs
+++ b/crates/nucleus-node/src/main.rs
@@ -3631,34 +3631,20 @@ impl NodeService for GrpcService {
         let pod_id_str = req.pod_id;
         let handle = pod_api::grpc_scoped_pod(&self.state, &md, &pod_id_str).await?;
 
-        let built = pod_receipt::build(&handle).await.map_err(|e| match e {
-            pod_receipt::ReceiptError::NotExited => Status::failed_precondition(e.to_string()),
-            pod_receipt::ReceiptError::NoExitReport(_) => Status::not_found(e.to_string()),
-            pod_receipt::ReceiptError::Malformed(_) => Status::internal(e.to_string()),
-        })?;
+        let built = pod_receipt::build(&handle, &self.state.authority)
+            .await
+            .map_err(|e| match e {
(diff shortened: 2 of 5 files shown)


brandon-coproduct:
Diagnosis of the two reds here, since one of them found a real hole and it is not in your diff.

**`exemplar metrics ratchet`** — `permissive_verify regressed 13->14 (lower is better)`. That is yours: the branch adds a permissive verify site. Either lower it or move the ratchet deliberately.

**`Manifest Guards`** — `FAIL scripts/check-verify-strict.sh — harness says green, the script says RED`, on `#16 ring-Ed25519 gate FAILED: cofactored ring::signature::ED25519 verify on a production path`.

The gate is right — `ring::signature::ED25519` is the cofactored verifier, it accepts the small-order identity triple, and `ring` has no strict variant. `crates/portcullis/src/manifest_registry.rs:814` is the test that demonstrates the attack, so a production-path one is the regression it exists to catch. Signing on `ring` (`Ed25519KeyPair`) is fine and the scan does not match it; it is the *verify* side that must be `verify_strict`.

**What your PR exposed, which is the part worth keeping.** `check-verify-strict.sh` carries two gates and `cargo xtask allowlist-gates` ported one. On a clean tree that is invisible: `--parity` compared one exit status against one aggregate verdict, so a gate the port never took contributes green exactly like a gate that is satisfied. It took a PR writing the forbidden construct for the two to disagree.

#2868 makes parity per-gate and requires every verdict line a script prints to be claimed by some gate, so an unported gate reds with no violation anywhere in the tree — and adds the eighth gate. It needed two new shapes to express, both of which are why it was droppable in the first place: no allowlist at all (the script says so in as many words), and an empty population as the *passing* state, which the non-vacuity rule reads as "stopped watching".

Nothing was unguarded meanwhile — `ci.yml:178` still runs the script, which is what red here.

PR #2864, opened 2026-09-12
Every name nucleus gives gatehouse must resolve to the same gatehouse

brandon-coproduct:
`cargo xtask gatehouse-pin` checked that two named workflows agree. It now checks that the pins agree, and adds a second fact the pins never covered.

## The population is discovered, not listed

The two workflows were named in a constant, so the rule enforced was "these two agree" and not "the pins agree". A third workflow pinning `GATEHOUSE_REF` would have sat outside the gate entirely, and the gate would have stayed green while its own failure message claimed something wider. Same defect as a `head -1`: a check whose subject is the first N of a thing rather than all of it.

The refs are now found by walking `.github/`, so a new pin joins the population by existing. A malformed pin is an **error rather than a skip** — skipping is how `GATEHOUSE_REF: main` leaves the population with nothing saying so.

## A step can run a gatehouse no pin names

`GATEHOUSE_REF` pins a gatehouse to **build**. The action does not have to build one. With `bin-dir` empty — **its default** — it downloads the release named by its own `version` input (`v0.1.0`) from `binaries`, and that release is a fourth gatehouse no pin in this repository mentions.

So a step that omits `bin-dir` runs a gatehouse chosen by an action default while the workflow two lines above pins a commit, and **both look pinned**. nucleus's one step sets it. The day a second is written the short way, the shadow lane compares a gate built from `GATEHOUSE_REF` against one downloaded from `v0.1.0` and reports agreement between them as a fact.

## A-19, and the exemption it frees

The uncovered entry read *"takes `--gatehouse <path>`; the probe needs a gatehouse checkout this script does not have"*. True of the conjunct that hashes `prelude/ci.writ` at the pinned ref — and of neither of the other two, which are decided from this tree before the flag is consulted. The same shape as `scoreboard-ratchet`, `check-dep-ceiling` and `check-wasm-closure` before it: **the obstacle named the gate's subject and never touched its detection.** `UNCOVERED_CEILING` 8 → 7.

`probe_xtask_partial` is what makes probing bare honest rather than convenient. `probe_xtask` refuses any gate CI calls with flags, correctly — probing a different command tests something CI does not run — so the licensed difference is paid for with two assertions `probe_xtask` does not make:

* the **clean** bare run must be green, or a red under perturbation is measuring "could not look";
* the **perturbed** run must red **with an expected marker in its output**, not merely exit non-zero. Exit status alone cannot tell "the conjunct under test fired" from "the gate fell over on the way there", which is exactly what a bare invocation invites.

CI's flags are still asserted verbatim, so the day they move this stops being a licensed difference and says so.

Both perturbations are real defects, not contrivances: the ref skew is **the one that happened** (plan lane `4d42510`, shadow lane `7326bfa9`), and the dropped `bin-dir` is the action's own default.

## A finding from building it

Both probes first reported "changed the file not at all", and the perturbations were correct — they change the file standalone. They were defined 100 lines *below* the probe calls. A script is not a program: a function exists from the line that defines it, so `"$@" "$target"` ran a command that did not exist yet, bash said so on stderr the harness discards, and the probe saw an unchanged file.

**The no-op guard is what caught it**, doing exactly its job. Recorded because the other reading of two identical FAILs is that both perturbations were wrong.

## Verification — exit codes, not tails

| check | rc |
|---|---|
| `cargo xtask gatehouse-pin` (clean tree) | 0 |
| ref skew perturbation | 1, `the pins name different gatehouses` |
| `bin-dir` dropped | 1, `a step runs a gatehouse no pin names` |
| `GATEHOUSE_REF: main` | 1, `is not a 40-character commit sha` |
| `cargo test -p xtask` | 0 — 150 passed, 15 in this module |
| `cargo clippy -p xtask --all-targets -- -D warnings` | 0 |
| `cargo fmt -p xtask -- --check` | 0 |
| `scripts/check-gates-can-fail.sh` | **0** — 39 probed, 7 uncovered (ceiling 7), 0 unaccounted |

`shellcheck -S error` reports the same 8 pre-existing findings on this branch as on main — none added here; #2825 fixes them.


brandon-coproduct:
Rebased onto main (`1f2547bb`) after #2862 landed. **Three** conflicts, all in `check-gates-can-fail.sh`, all unions:

1. **probe call sites** — main gained #2825's `probe check-dep-ceiling.sh`, this branch adds two `probe_xtask_partial gatehouse-pin`. Both kept.
2. **`UNCOVERED_CEILING`** — main is now **6** (#2825 took it 8 → 7 → 6). This branch frees `gatehouse-pin`, so **6 → 5**, with both histories' reasoning kept rather than one side's.
3. **the accounting regex** — main matches `probe_xtask(_generated)?`, this branch `probe_xtask(_partial)?`. Taking either alone **drops the other form**, and a probed gate then reads as unaccounted. Merged to `(_generated|_partial)?`.

That third one is the same trap #2870's rebase hit, and it will recur: whichever of #2864/#2870 lands second has to widen the alternation again (`_flagged` is #2870's). A both-sides-additive conflict where each side edited the *same predicate* needs union of the alternation, not union of the lines — worth knowing before resolving it the fast way.

Re-verified after the rebase: `gatehouse-pin` **0**, `cargo test -p xtask gatehouse_pin` **0** (15 passed), clippy `-D warnings` **0**, `fmt --check` **0**, `shellcheck -S error` **0**, and **`SUITE_RC=0`** — 42 probed, 5 uncovered (ceiling 5), 0 unaccounted, with both new probes green:

```
ok  xtask gatehouse-pin — RED on two workflows building gatehouse at different commits, GREEN when restored
ok  xtask gatehouse-pin — RED on a step falling back to the action's downloaded default, GREEN when restored
```

brandon-coproduct:
Rebuilt on main (`478c09be`) after #2871 landed, and **dequeued first** — a branch in the merge queue cannot be pushed to:

```
remote: error: GH006: Protected branch update failed
remote: - A pull request for this branch has been added to a merge queue. Branches that
remote:   are queued for merging cannot be updated.
```

## Why it needed rebuilding rather than rebasing

#2871 took `UNCOVERED_CEILING` 6 → 5. This branch still set 5 while removing one *more* entry, so the merge result would have been **4 entries against a ceiling of 5** — a certain failure, confirmed mid-rebase rather than assumed. It was at queue position 5 and would have burned a build slot to discover it.

Replaying four commits over the new main gave **seven conflict regions** in `check-gates-can-fail.sh`, every one the "both sides added a helper" shape — and among them the accounting *predicate*, where "take both lines" is silently wrong rather than merely noisy (gatehouse F-137). Rebuilding from main and re-applying four known blocks is deterministic; resolving seven regions by hand is not.

## One live hole found on the way

Main's predicate reads `probe_xtask(_generated)?`, and main gained `probe_xtask_flagged` in #2868 — so **a gate probed solely by the flagged form already reads as UNACCOUNTED on main**, and is correct there only because the one user is also probed bare (F-142). This branch adds `_partial`, which would be the second such hole.

Derived to `^probe_xtask[a-z_]*` instead — which is what #2870 argues for, so the two branches now **converge rather than conflict**.

| | |
|---|---|
| both `gatehouse-pin` probes | green |
| `gatehouse-pin` · `cargo test -p xtask gatehouse_pin` | 0 · 0 (15 passed) |
| clippy `-D warnings` · fmt · shellcheck | 0 · 0 · 0 |

Entries and ceiling counted by hand as well as by the gate: 4 and 4.

**#2870 and #2876 need the same treatment** — both still set ceiling 5 with a list one shorter than main's. Recorded as gatehouse F-149: a shrink-only ratchet whose value every contributor must decrement serialises them, and the serialisation is visible in the number but hidden in everything the branches touched on the way there.

PR #2865, opened 2026-09-12
A snapshot base cannot carry a mounted writable scratch — a plan correction

brandon-coproduct:
The "nucleus builds nucleus" plan described making pods snapshot-eligible as **"two one-line policy changes"**: turn on `--firecracker-api-boot`, and stop `scratch_for_pod` auto-provisioning a writable disk.

The first is one line. **The second is not a policy question at all**, and the reasoning already sitting in `snapshot.rs` is better than the plan's.

## What I got wrong

`clone_safety` returns `WritableScratchAttached` as its *first* refusal, so every jailed pod is unsnapshottable today. I read that as a default to flip. Its doc comment states the real constraint:

> *Scratch is per-pod and writable. A restored clone inherits the base's in-memory ext4 state for it, so two clones pointed at one file corrupt each other, and giving each a fresh file at the same in-jail name leaves the guest's cached metadata describing a filesystem that is no longer there.*

The obvious workaround **is the second of those cases**. Firecracker requires disk backing files at the same *relative* path to the restoring process, and each pod has its own jail — so `/scratch.ext4` can be a different host file per pod at the same relative path. That looks like it solves the sharing problem, and it is exactly the corruption the comment already names.

## Checked upstream rather than assumed

| source | says |
|---|---|
| Firecracker snapshot design | sharing is for memory pages and **read-only** disks |
| firecracker-containerd #759 | EXT4 inode-checksum errors after loading a snapshot; restored disk state inconsistent |
| Firecracker #4014 | drives **cannot** be reconfigured for a snapshot-booted VM — no "attach a different disk at restore" escape |
| Firecracker #5795 | block-level CoW overlay with dirty-bitmap tracking, which *would* solve this, is an open **feature request** — not in the pinned v1.16.1 |

## The candidate design, marked untested

**Attach the scratch but defer the MOUNT past the barrier.** The drive must exist in the config at snapshot time or the paths won't match; but if the guest never mounted it, the snapshot holds only virtio-blk queue state and no ext4 metadata. Mounting becomes part of *personalization*, which is where the barrier already puts `FETCH_POD_SPEC`.

Two things it needs that don't exist: `clone_safety`'s predicate is attachment-based and would have to become mount-based (and the host can't see a mount — the guest must say so, as `SNAPSHOT_READY` already does for "I've asked for nothing"); and **validation**, because "device state survives a substituted backing file of identical geometry" is reasoning, and #759 is what that reasoning failing looks like.

## Impact on sequencing

M2 does not depend on this. Command-in (#2861), result-out (#2862) and host-signing (#2863) are independent of whether the base is snapshot-restorable — **a pod that cold-boots produces the same receipt as one that resumes.** Finish M2, then validate the hypothesis on real hardware before writing any of it into a design.

### DIFF
--- a/docs/snapshot-scratch.md
+++ b/docs/snapshot-scratch.md
@@ -1,0 +1,78 @@
+# A snapshot base cannot carry a mounted writable scratch
+
+**2026-09-12.** A correction to the "nucleus builds nucleus" plan, which
+described making pods snapshot-eligible as *"two one-line policy changes"*:
+turn on `--firecracker-api-boot`, and stop `scratch_for_pod` auto-provisioning
+a writable disk. The first is a one-line change. **The second is not a policy
+question at all**, and the reasoning already in `snapshot.rs` is better than
+the plan's.
+
+## What the plan got wrong
+
+`clone_safety` returns `WritableScratchAttached` as its **first** refusal, so
+every jailed pod is unsnapshottable today. The plan read that as a policy
+default to flip. It is not — the doc comment states the actual constraint:
+
+> Scratch is per-pod and writable. A restored clone inherits the base's
+> in-memory ext4 state for it, so two clones pointed at one file corrupt each
+> other, and giving each a fresh file at the same in-jail name leaves the
+> guest's cached metadata describing a filesystem that is no longer there.
+> Both are filesystem corruption arriving later and elsewhere.
+
+The obvious workaround does not work, and is the *second* case above.
+Firecracker requires that disk backing files be at **the same relative path**
+to the restoring process (`docs/snapshotting/snapshot-support.md`), and because
+each pod has its own jail, `/scratch.ext4` can be a different host file per
+pod at the same relative path. That looks like it solves the sharing problem.
+It does not: the guest's cached ext4 metadata came from the snapshot and now
+describes a filesystem that is not underneath it.
+
+## Upstream agrees, and has the scars
+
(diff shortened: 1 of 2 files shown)


brandon-coproduct:
The fmt shadow red here is **informational and correct**, and the error line that should have explained it was empty. Diagnosis, since the log makes it hard:

```
could not look: ProofNotAtCheckpoint { proof_size: 693, trusted_size: 694 }
##[error]gatehouse: could not verify the receipt against the log:
```

Two adjacent lines, unconnected. `gatehouse.sh` captures `VOUT=$("$GATE" receipt verify …)` — stdout — and `gate receipt verify` reports on **stderr**, so the one `die` designed to say why interpolated nothing. #2869 fixes that with `2>&1`.

The underlying cause is a **race**, not anything in your diff: the inclusion proof arrives with the receipt POST and the trust root comes from a second request, so a receipt landing between the two puts the checkpoint one ahead of the proof — 693 vs 694. Thirteen PRs are running shadow gates against one control plane right now, so it is not rare, and it gets likelier with adoption rather than rarer.

A proof at size *n* should verify against a checkpoint at *n+k* — a prefix is a prefix — so the fix belongs in gatehouse's verifier rather than here. Recorded as gatehouse F-121. The check's own name says `red = could not look, never disagreement`, and that is exactly what this is: **not a blocker for this PR.**

PR #2866, opened 2026-09-12
A scratch blocks a snapshot base when MOUNTED, and the host can see which

brandon-coproduct:
M3. `clone_safety`'s first refusal was `writable_scratch` — *is a writable non-root drive attached* — and `scratch_for_pod` auto-provisions one for every jailed pod. So **every jailed pod was unsnapshottable.**

Measured 2026-09-12 (`scripts/experiments/snapshot-deferred-mount.sh`, Firecracker v1.16.1 aarch64/KVM): the corruption that refusal exists for is about a **mounted** scratch. Attached-and-never-mounted leaves virtio-blk queue state in the snapshot and no ext4 metadata — two clones restored against fresh images of identical geometry both fsck clean, wrote only their own file, and logged zero `EXT4-fs error`.

## The host looks rather than asking

The obvious way to get a mount-based predicate is to ask the guest: it does the mounting, and `SNAPSHOT_READY` is already a guest report. **That would be the wrong choice.** A base is restored by *other* pods, so a guest that lies about its mount state doesn't corrupt itself — it publishes a base that corrupts everyone who restores it. A claim with that blast radius must not rest on the claimant.

ext4 records it in the superblock, so the host reads it:

```
fresh, never mounted     Mount count: 0    Last mounted on: <not available>
mounted and unmounted    Mount count: 1    Last mounted on: /mnt
```

`dumpe2fs -h` — no new dependency, since `provision_pod_scratch` already shells to `mkfs.ext4` to *create* the image.

## Unknown refuses

A superblock this code can't parse is `MountState::Unknown` and blocks the base, sitting beside `Mounted` rather than beside `NeverMounted`. The costs aren't symmetric: **a wrong refusal is a cold boot; a wrong certification corrupts every clone of the base.**

`SnapshotInputs` now carries the scratch *path*, not a bool — the question is about the image's superblock, so the caller needs the file rather than a flag saying one exists. One expression of the fact, not two.

## Tests

Seven, including the two that matter in opposite directions: an attached-unmounted scratch is now cloneable, and a mounted one still isn't. Parsing is separated from `dumpe2fs` so it's tested on macOS, where e2fsprogs is absent.

## What remains

`guest-init` still mounts `/work` at boot, so the count is 1 by the barrier. Deferring that mount past the barrier is the guest-side half — **and this gate refuses until it happens**, which is the right order: the gate before the behaviour.

### DIFF
--- a/crates/nucleus-node/src/firecracker_config.rs
+++ b/crates/nucleus-node/src/firecracker_config.rs
@@ -855,10 +855,10 @@ impl FirecrackerConfig {
             smt: self.machine_config.smt,
             // A scratch drive is the writable, non-root one. `is_read_only` is the property that
             // matters, not the drive's name, because a name is a convention and this is not.
-            writable_scratch: self
-                .drives
-                .iter()
-                .any(|d| !d.is_root_device && !d.is_read_only),
+            scratch_path: self.drives.iter().find_map(|d| {
+                (!d.is_root_device && !d.is_read_only)
+                    .then(|| std::path::PathBuf::from(&d.path_on_host))
+            }),
         }
     }
 
--- a/crates/nucleus-node/src/firecracker_config_tests.rs
+++ b/crates/nucleus-node/src/firecracker_config_tests.rs
@@ -81,14 +81,16 @@ fn a_read_only_data_image_does_not_count_as_writable_scratch() {
     assert!(
         !cfg(&with_data)
             .snapshot_inputs(nucleus_spec::vmm_version::PINNED)
-            .writable_scratch,
+            .scratch_path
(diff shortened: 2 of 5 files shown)


brandon-coproduct:
`Clippy` here is one real error, not a gate problem:

```
error: this boolean expression can be simplified
  --> crates/nucleus-node/src/firecracker_config_tests.rs:82:9
error: could not compile `nucleus-node` (bin "nucleus-node" test) due to 1 previous error
```

`clippy::nonminimal_bool` under `-D warnings`. One line, in your diff. Nothing else on this PR is red.

PR #2867, opened 2026-09-12
/work is mounted past the snapshot barrier, and earlier does not compile

brandon-coproduct:
The guest-side half of M3. `guest-init` mounted `/dev/vdb` at `/work` with the other mounts, ~50 lines *before* it announced the snapshot barrier.

Mounting ext4 **writes to its superblock** — mount count goes to 1, the last-mounted path is recorded — so a base snapshotted at the barrier already carried one pod's filesystem metadata. #2866's host-side check refuses exactly that, so without this it would refuse forever.

The barrier's own comment already had the principle:

> *Announce the barrier before asking for anything. After the first fetch below this VM is one particular pod.*

Mounting a per-pod filesystem **is** that, and it was happening first.

## A token, not a convention

`mount_work` takes an `identity::PastBarrier`, which only `identity::barrier` can construct. Mounting the scratch before the barrier isn't a rule someone has to remember — **it does not compile.**

## The first version of that was decorative, and a test caught it

`PastBarrier` started in `main.rs` beside `mount_work`. Rust's field privacy is **module-scoped**, so the marker was constructible by the very code it was meant to constrain. I perturbed the call site to mount before the barrier and it **compiled straight through**.

Moved to `identity.rs`, the same perturbation fails:

```
error[E0603]: tuple struct constructor `PastBarrier` is private
```

The module boundary is the whole mechanism, and it's worth saying out loud because the type looks identical either way. A marker in the same module as its user constrains nothing.

## The guest is not being believed

The host verifies the *outcome* in the superblock (`snapshot::mount_state`, #2866) rather than trusting this. The guest change is the behaviour that makes the host's check pass — not a claim the host accepts.

### DIFF
--- a/crates/nucleus-guest-init/src/identity.rs
+++ b/crates/nucleus-guest-init/src/identity.rs
@@ -190,6 +190,32 @@ pub struct BrokerCapability {
 /// returns `Ok(())` anyway: announcing a barrier is not a request for anything, so failing to
 /// announce it must never stop a pod from starting. The cost of a host that never hears it is
 /// that the VM is refused as a snapshot base, which is the correct answer.
+/// Proof that the snapshot barrier is behind us.
+///
+/// The field is private and this type lives HERE, not beside its user. That is
+/// load-bearing: Rust's field privacy is module-scoped, so a marker defined in
+/// the same module as the code it constrains can be constructed by that code
+/// and constrains nothing. The first version of this was in `main.rs` and a
+/// perturbation test compiled straight through it.
+///
+/// Defined here, the only way to hold one is [`barrier`] — which announces.
+pub struct PastBarrier(());
+
+/// Announce the snapshot barrier and return the proof that it happened.
+///
+/// Best-effort by design: a host that never hears it simply never records the
+/// barrier, and the only consequence is that this VM cannot serve as a base. A
+/// VM with no workload API has no barrier to announce and is already unusable
+/// as one — either way nothing is left to defer, so the token is issued.
+pub fn barrier(port: Option<u32>) -> PastBarrier {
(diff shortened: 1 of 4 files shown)


brandon-coproduct:
`a real nucleus pod boots and is enforced` is a **real failure in this branch**, not infrastructure and not a stale gate.

The node comes up fine and then the pod create dies 30 seconds later:

```
  [OK] nucleus-node is answering on https://127.0.0.1:8080
Error: the node refused to create a pod: error sending request for url (https://127.0.0.1:8080/v1/pods)
```

and the guest-side log is a loop of

```
vsock bridge connection error: Connection refused (os error 111)
vsock bridge shutting down for pod 7f8f59eb-7955-4449-af64-1680f9d3f044
```

So every preflight passes — `/dev/kvm`, `/dev/vhost-vsock`, firecracker, guest kernel, rootfs, node+secrets — the node answers, and the pod's vsock bridge then cannot connect. Given the branch defers the work mount, the shape fits: something the bridge needs is no longer there when it reaches for it.

**It is this branch.** `quickstart-boot` on the last twelve runs:

| branch | result |
|---|---|
| `main` ×4 | success |
| `gh-readonly-queue/…` (#2868, #2865, #2861, #2862, #2825) | success |
| `test/receipt-composition` | success |
| **`ci/defer-work-mount`** | **failure** |

Full job: https://github.com/coproduct-opensource/nucleus/actions/runs/34677462562/job/103509773705 — the failing step is `Boot a real pod and assert what the guest did`, `06:11:03Z → 06:18:58Z`.

Leaving it to you since it is in the diff's own subject; flagging so it is not read as the shadow-gate noise on the other PRs.

PR #2870, opened 2026-09-12
A job routed at a runner that does not exist waits exactly like a busy one

brandon-coproduct:
Two things go wrong at `runs-on:` and neither announces itself. An expression that evaluates to the **empty string**, and a label naming a **pool nobody deployed**, both present as a job sitting in `queued`.

That is also what GitHub shows for a job waiting on a busy pool — and this repository shares **eight** build machines between its pull requests and its merge queue, so *waiting for a runner* is the ordinary state and an unroutable job hides inside it perfectly.

Measured 2026-09-12 03:46Z while writing this: **30 in-flight `pull_request` runs, 0 `merge_group`**, with the merge queue's head entry #2826 holding its CI run in `queued` for 24 minutes. It drained eight minutes later, so that instance was starvation and not a routing fault — **but nothing in the picture said which**, and `check_response_timeout_minutes = 360` with `max_entries_to_build = 1` means the wrong reading costs six hours of a queue that admits one entry at a time.

## Both questions are decidable from the tree

| conjunct | rule |
|---|---|
| **routable** | an expression that can yield *nothing* must be guarded by an `if:` refusing that case |
| **declared** | every `vars.*` a `runs-on:` reads must be a runner variable this repository has an opinion about, and every self-hosted label it can produce must be a pool `manager.toml` declares or a listed exception with a reason |

Exactly one site relies on the first — `quickstart-boot.yml`'s aarch64 job, `runs-on: ${{ vars.ARM64_METAL_RUNNER }}` paired with `if: … && vars.ARM64_METAL_RUNNER != ''` — and it is **correct**. The gate makes that pairing load-bearing instead of incidental, and a unit test pins that the guard is *found*, not merely that the gate passes.

The variable *values* live in GitHub's repository variables and a checkout cannot read them. What a checkout can answer, and nothing asked, is whether anybody decided on this runner at all.

**148 sites**, counted independently before the gate was written and again by it.

## Two parser corrections, both found by running it

- `aeneas-ifc-scoped.yml`'s `runs-on` is a ternary, and taking every quoted string made `needs.scope.outputs.relevant == 'true'` offer `'true'` as a runner label — **red on a correct workflow**. A literal on the right of `==` or `!=` is an operand of the condition, never a value the expression produces.
- a bare `runs-on: ubuntu-latest` has no quotes, so the literal scan skipped it — **79 of the 148 sites**. The gate would have printed *"all routable"* while checking 69. That is the defect it exists to refuse, in its own population.

The second one is worth how it surfaced: A-19's third perturbation **exited 1 with no FAIL line printed**, so the red came from an unrelated conjunct. Exit status alone would have passed that probe — it confirmed the gate can fail, not that it can fail at the thing being probed.

| | rc |
|---|---|
| `fly-pools` clean | 0 — `148 runs-on site(s), all routable` |
| guard removed from the unfallbacked site | 1 |
| a job routed at `vars.CI_HEAVY_RUNNER` | 1 |
| a literal label no pool declares | 1 |
| `cargo test -p xtask` | 0 — 146 passed, 10 in this module |
| `clippy --all-targets -D warnings` · `fmt --check` · `actionlint` · `check-ci-spec.sh` | 0 · 0 · 0 · 0 |
| `scripts/check-gates-can-fail.sh` | **0** — 38 probed, 0 FAIL, 0 unaccounted |

`UNCOVERED_CEILING` untouched at 8 — the new probe is a second perturbation for a subcommand already counted, not a freed exemption, so this does not conflict with #2825 or #2864.


brandon-coproduct:
Added a second commit, `13f2b50`, from a red found while surveying: **#2835's `Gates must fail on their own subject` is a misattribution, four times over.**

`probe_xtask` measured each gate perturbed then restored, and read a red restore as *"the perturbation left something behind"*. It cannot mean only that. On #2835, `xtask scorecard` is already red for its own reasons — a ratchet floor the base does not meet, visible two checks away as `Manifest Guards: FAIL: 1 scorecard finding(s)` — so every probe of it reported `still failing (exit 1) after restore`. Four failures, all naming the one thing that did not happen, sending any reader hunting for a perturbation that does not clean up.

The baseline is now measured **before** anything is touched. Already red, the probe says exactly that and stops: a gate failing for its own reasons decides nothing under perturbation. The restore message can now only mean what it says, and says so.

Same assertion `probe_xtask_partial` carries in #2864, reached from the opposite direction — there because a bare invocation might not reach the conjunct under test, here because the tree might be red before the probe starts. Both are the rule that **a comparison needs a known left operand.**

Falsified before green: with `mediation/net` pinned below its size, the probe prints `already red (exit 1) BEFORE any perturbation` instead of blaming the restore. Suite re-run after the change: exit **0**, 38 probed, 0 FAIL, 0 unaccounted.

brandon-coproduct:
Third commit, `8ff20d16`: the **same baseline check on `probe`**, the shell half — which covers **30 of the harness's 38 probes**.

`probe_xtask` got it in the previous commit because that is where the symptom appeared (#2835's `xtask scorecard`). `probe` had the identical code, the identical message and the identical gap, and nothing made it immune — it had simply not yet met a gate that was red on arrival. Fixing only where the symptom showed would have left the majority of the harness able to reproduce it.

Falsified in that half specifically: emptying `scripts/verify-strict-allowlist.txt` makes `check-verify-strict.sh` red on arrival, and the probe now prints

```
FAIL  check-verify-strict.sh — already red (exit 1) BEFORE any perturbation.
```

instead of blaming the restore, with `covered` staying 0 — which is right, it covered nothing.

Suite re-run on this commit: **`SUITE_RC=0`**, 38 probed, 0 FAIL, 0 unaccounted.

Recorded as gatehouse F-125. Related: **F-123** — while surveying, #2835/#2838/#2840's `scorecard` red turned out to be a ratio whose *denominator* grew (`22/80` pinned, `22/81` measured, `discharged` unchanged at 22) while the gate reported it as the numerator falling. That is the same class as this one level up, and it is why the misattribution mattered enough to chase.

brandon-coproduct:
Fourth commit, `49fbb51`: **three gates run in CI on every push and the accounting had never seen them.**

`XTASK_GATES` was derived by globbing `.github/workflows/*.yml`. A gate does not have to be named in a workflow to run in CI — it can be reached through a script a workflow runs, and three are:

```
scripts/check-inert-authority.sh   exec cargo run -q -p xtask -- inert-authority "$@"
scripts/check-law-mechanisms.sh    exec cargo run -q -p xtask -- law-mechanisms "$@"
scripts/check-kani-proof-count.sh  cargo run -q -p xtask -- kani-coverage
```

None appeared in the domain, none had an exemption row, and **the accounting still printed `0 unaccounted`** — because the derivation never looked where they live.

That is this section's own thesis one level out. Its comment opens by saying the shell loop globs `scripts/check-*.sh`, so a gate that is not a shell script was never in the domain being searched, and the fix was to derive the xtask half from the workflows. The workflows are not the whole of CI either. Composite actions are now in scope for the same reason.

**Domain 14 → 16.** The three are accounted for by `SHIM_COVERED`, and a row there is a claim that gets *checked* rather than asserted:

- the named script must exist and must **actually invoke** the named subcommand, so a shim that is rewritten or a call that moves stops being covered by that line instead of quietly staying exempt;
- a row naming a subcommand nothing in CI invokes at all is a gate ranging over nothing, which reads exactly like live coverage.

`so` stays out of the domain, which it would not without the `^[^#]*` prefix: the prose *"added to xtask -- so including the lock left the 46 minutes"* in `coverage-matrix.yml` is a comment. Same trap `probe()` already records for shell gates, verified after extending the glob to `scripts/` where it had a second chance to bite.

Suite on this commit: **`SUITE_RC=0`**, 38 probed, 0 FAIL, 0 unaccounted.

Recorded as gatehouse **F-126**. The recurring form — five times this session (F-108, F-113, F-119, F-126) — is that **a coverage claim is only ever as wide as the enumeration behind it, and the enumeration is the part nobody re-reads.**

brandon-coproduct:
A-19 for the two new conjuncts, driven end to end rather than by exit status. Perturbation: the shim renamed to call `xtask -- inert-authority-renamed`, **committed** — the first attempt exited 1 from the dirty-tree guard and never reached the accounting, which is the trap F-119 records.

```
FAIL  check-inert-authority.sh — already red (exit 2) BEFORE any perturbation.
FAIL  SHIM_COVERED names xtask inert-authority, which nothing in CI invokes at all.
accounting: 36 gate script(s) found, 36 probed, 8 listed uncovered, 1 unaccounted
    xtask inert-authority-renamed
FAILED: 4 problem(s). A gate that cannot fail is not a gate.
```

Three things fired and each is the one under test:

1. **the stale-row conjunct** — a `SHIM_COVERED` row for a subcommand nothing invokes is a gate ranging over nothing, and it reads exactly like live coverage;
2. **the domain extension** — `inert-authority-renamed` is reachable *only through a script*, and it lands in `UNACCOUNTED`. Before this commit it would have been invisible, and the accounting would have printed `0 unaccounted`;
3. **the baseline check** from `8ff20d16`, on a real case — exit 2 is "unrecognized subcommand", so the gate is red on arrival and the probe says so instead of blaming the restore.

Tree restored to `49fbb51`, clean. On the unperturbed commit: **`SUITE_RC=0`**, 38 probed, 0 FAIL, 0 unaccounted.

brandon-coproduct:
Rebased onto main now that #2825 has landed (`fa0479a0`). Two conflicts, both in `check-gates-can-fail.sh`, both resolved as unions:

1. **the probe call sites** — main gained #2825's `probe_xtask_generated scoreboard-ratchet`, this branch adds `probe_xtask fly-pools`. Both kept.
2. **the accounting regex** — main matches `probe_xtask(_generated)?`, this branch `probe_xtask(_flagged)?`. Taking either side alone would have **dropped the other form from the accounting**, so a probed gate would have read as unaccounted. Merged to `probe_xtask(_generated|_flagged)?` with a comment saying there are now three forms and why listing one is not enough.

That second one is worth naming: a both-sides-additive conflict where each side edited the *same* predicate is the case where "take both" has to mean union of the alternation, not union of the lines.

Re-verified after the rebase: `fly-pools` **0**, `cargo test -p xtask fly_pools` **0** (10 passed), `shellcheck -S error` **0** on the whole harness — main now has zero too, since #2825 carried those fixes.

brandon-coproduct:
Fifth commit, `954f16e6`: **`lean-action-builds` was never uncovered, and the row that would have said so was wrong.**

### The exemption named a different program's obstacle

It sat in `UNCOVERED` as *"needs a Lean toolchain to reach its verdict"*. `crates/xtask/src/lean_action_builds.rs` contains **no `Command` and no `process::`** — it reads `.github/workflows/*.yml` and parses YAML with serde_yaml — and it **reaches no verdict**, because its own header says what it is: *"read explicit Lean-action build inputs for the library-coverage gate"*. The gate is `check-lean-libs-built.sh`, which calls it at line 50 and is itself probed twice.

Fifth instance of the shape `scoreboard-ratchet`, `check-dep-ceiling`, `check-wasm-closure` and `gatehouse-pin` each turned out to be — and the sharpest. The other four named a **real** obstacle and stopped short of the detection half; this one named an obstacle belonging to a different program. Moved to `SHIM_COVERED`; **`UNCOVERED_CEILING` 6 → 5**.

### Moving it exposed a bug in SHIM_COVERED's own check, from one commit ago

The row is meant to be a **checked** claim — the named script must actually invoke the named subcommand — and the check was

```
grep -qE "xtask -- ${sub}([[:space:]]|\"|$)"
```

`check-lean-libs-built.sh` calls it as `$(cargo run -q -p xtask -- lean-action-builds)`, so the name is followed by `)`, which is none of those. **The check reported a live, correct row as false.** The three existing rows happened to be followed by a space (`-- inert-authority "$@"`) or end-of-line, so nothing had exercised the other syntax.

That is this session's recurring shape *inside the checker written to catch it*. Widened to any non-word boundary; all four rows verified true, then falsified by renaming the call.

The durable form, recorded as gatehouse F-135: **a verification rule is exercised only by the members it has, so adding a member is the cheapest test of the rule** — and a rule whose members all share one incidental shape has never been tested at all.

`shellcheck -S error` **0**; **`SUITE_RC=0`** — 41 probed, 5 uncovered (ceiling 5), 0 unaccounted.

brandon-coproduct:
Sixth commit, `f8c4d255`: **derive which probe helpers count, instead of listing them.**

The accounting asked `grep -qE "^probe_xtask(_generated|_flagged)?…"` to decide whether a gate is probed. A list is the wrong shape for that predicate: **every branch that adds a helper edits the same line**, so two such branches conflict there — and "take both sides" has to mean union of the *alternation*, not union of the lines. Resolved the fast way it silently drops a sibling's form, and a gate that **is** probed then reads as unaccounted.

**Three rebases hit this in one session** — #2864, this branch, and this branch again — each needing the distinction spotted by hand.

And the alternation that survived here named `_flagged`, which is defined on **#2868 and not on this branch**. Harmless at runtime, since the alternative never matches — but it is a coverage claim wider than the code, inside the file whose purpose is to refuse exactly that. I wrote it during a rebase resolution, by unioning two lists without checking that either side had both members.

Fixed by changing the shape rather than by being more careful: `^probe_xtask[a-z_]*` derives the family. Every helper lives in this file and is reviewed with it, so matching by name costs nothing the list was buying and removes the conflict class.

Falsified: renaming a probe's helper makes its gate stop matching, which is the UNACCOUNTED path — so the predicate still refuses what it exists for. Both existing forms still match.

**`SUITE_RC=0`** — 41 probed, 5 uncovered (ceiling 5), 0 unaccounted. `shellcheck -S error` 0.

Recorded as gatehouse F-137. The general form: *an enumeration every contributor must edit is a merge conflict with a silent wrong answer, and deriving the set costs nothing the list was buying.*

---

Heads-up on this branch, #2864 and #2871: all three set `UNCOVERED_CEILING` 6 → 5, so the first to merge ejects the other two as DIRTY. That conflict is safe to resolve — `6 − (gates freed by both)` is arithmetic, not a silent drop — but it will need a rebase each time.

brandon-coproduct:
Merged main in (`033979de`) and **dequeued first** — a branch in the merge queue cannot be pushed to (`GH006: … Branches that are queued for merging cannot be updated`).

#2871 and #2868 landed while this sat at queue position 5. Two conflicts, both in `check-gates-can-fail.sh`:

1. **two perturbation functions** at the same spot — this branch's `perturb_runs_on_undeclared_var` and main's `perturb_unported_shell_gate`. Unrelated and additive; both kept.
2. **the ceiling's comment block** — main records freeing `check-wasm-closure`, this branch records freeing `lean-action-builds`. Both histories kept, and the **value moved 5 → 4**, which is the arithmetic rather than either side: main's list is 5 long and this removes one more.

The second is why it had to be touched at all. Queued as it was, it would have merged to **4 entries against a ceiling of 5** and failed — `Uncovered: 4 (ceiling 5)` — after waiting for a build slot to discover it.

| | |
|---|---|
| `fly-pools` · `shellcheck -S error` | 0 · 0 |
| derived `^probe_xtask[a-z_]*` predicate | intact |
| all four `SHIM_COVERED` rows | present |

Entries and ceiling counted by hand as well as by the gate: 4 and 4.

Recorded as gatehouse **F-149** — a shrink-only ratchet whose value every contributor must decrement serialises them, and the serialisation is visible in the number but hidden in everything the branches touched on the way there.

PR #2876, opened 2026-09-12
policy-gate gets a probe; "needs a real amendment" needed a copy of a file

brandon-coproduct:
The sixth exemption this session whose stated obstacle named the gate's **subject** and never touched its **detection**.

It read *"runs ck-kernel admission on a manifest amendment; needs a real amendment"*. An amendment is a second manifest. `policy_gate` reads two TOML files and a text file and calls `gate_manifest_amendment` in `GateMode::Preflight`, which builds the kernel `with_skip_for_testing()` — **signatures deliberately skipped**, no witness bundle, no network, no governance ceremony to arrange.

From the committed `PolicyManifest.toml` alone:

| | |
|---|---|
| `base == candidate` | ACCEPTED, exit **0** |
| +1 entry in `network_allow` | exit **1**, `CapabilityNonEscalation: Capability escalation: ["network_allow: +[evil.example.com]"]` |

which is the kernel's whole point. `UNCOVERED_CEILING` **6 → 5**.

## A second generated-input slot, because one wasn't enough

CI feeds this gate two files the job writes first — a base manifest and a changed-files list. `probe_xtask_generated` substituted exactly one path and refuses `generated == local_path` by design, so the pair was inexpressible.

The obvious workaround is worse than it looks: have the generator write the second file at CI's own path in the repo root. That leaves it behind, and **the dirty-tree guard at the top of this same script refuses to start on an untracked file** — so the probe would run once and then block every subsequent run of the suite. A probe whose cost is that the next run cannot start is not a probe.

So the slot is real, and it carries the rule that found it: **both generated paths must be absolute**, checked before anything is written, with the reason in the failure text. The existing single-input call site is untouched — the new parameters are optional. The second input is checked with `-f` not `-s`: a changed-files list is legitimately empty when an amendment touches no `may_not_modify` path.

## Then the parity check caught itself

The flags extractor scanned **one line at a time**, and `ck-admit.yml` spells this invocation across four backslash continuations — so what the check believed CI ran was `\`:

```
FAIL  xtask policy-gate — CI invokes it as: \
```

It fails **safe**, refusing rather than passing, which is exactly why nothing had noticed: every probe that existed was a single-line invocation. Still wrong — it refuses a correct probe while asserting CI runs something it does not, and a parity check whose left operand is a line-continuation character is not comparing anything. Continuations are now joined before matching, in both extractors, with whitespace squeezed; the existing single-line cases verified unchanged.

## And a third commit, because I got the ratchet half right

The first commit lowered the ceiling 6 → 5 and **left `policy-gate` in the UNCOVERED list**. The suite said so — `Uncovered: 6 (ceiling 5)`, `FAILED: 1 problem(s)`. The removal was an exact-string replace including a run of spaces, the file had thirteen where I wrote twelve, it matched nothing and returned the string unchanged, and I did not check that it had.

That is the **no-op-edit shape this harness exists to catch**, in the edit that was adding a probe to it. Fixed by regex with the count asserted 1 → 0 and the list counted against the ceiling by hand. The ratchet caught it within one commit, which is what "exact in both directions" is for.

| | rc |
|---|---|
| `policy-gate` unperturbed · escalation · restored | 0 · **1** · 0 |
| `scripts/check-gates-can-fail.sh` | **0** — 41 probed, 5 uncovered (ceiling 5), 0 unaccounted |
| `shellcheck -S error` | 0 |

Recorded as gatehouse F-136, F-138, F-140.

### DIFF
--- a/scripts/check-gates-can-fail.sh
+++ b/scripts/check-gates-can-fail.sh
@@ -954,7 +954,7 @@ probe_xtask fly-pools ci/fly-runner/manager.toml \
 probe_xtask_generated scoreboard-ratchet scripts/exemplar-baseline.json \
     "a baseline claiming a score the tree does not have" \
     "--current scoreboard.json --baseline scripts/exemplar-baseline.json" \
-    "scoreboard.json" "$(mktemp -t scoreboard).json" \
+    "scoreboard.json" "$(mktemp "${TMPDIR:-/tmp}/scoreboard.XXXXXX").json" \
     gen_exemplar_scoreboard perturb_exemplar_baseline
 probe_xtask push-auth .github/workflows/clippy-ratchet.yml \
     "a CI push relying on the checkout's ambient credential" perturb_push_auth_strip
@@ -967,9 +967,9 @@ probe_xtask gate-budget .github/workflows/gatehouse-shadow.yml \
 probe_xtask_generated policy-gate PolicyManifest.toml \
     "an amendment the constitutional kernel must refuse" \
     "--base before.toml --candidate PolicyManifest.toml --changed-files changed.txt" \
-    "before.toml" "$(mktemp -t policy-base).toml" gen_policy_base \
+    "before.toml" "$(mktemp "${TMPDIR:-/tmp}/policy-base.XXXXXX").toml" gen_policy_base \
     perturb_policy_escalation \
-    "changed.txt" "$(mktemp -t policy-changed).txt" gen_policy_changed
+    "changed.txt" "$(mktemp "${TMPDIR:-/tmp}/policy-changed.XXXXXX").txt" gen_policy_changed
 
 probe check-line-ratchet.sh   "--strict" crates/portcullis/src/kernel.rs \
       "400 lines past the ceiling"            perturb_line_ratchet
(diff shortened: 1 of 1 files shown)


brandon-coproduct:
Merged main in (`af80b817`) and **dequeued first** — a branch in the merge queue cannot be pushed to (`GH006`).

#2871 landed while this sat at queue position 8, freeing `check-wasm-closure` and taking the ceiling 6 → 5. One conflict: the ceiling's comment block, where main records that change and this branch records freeing `policy-gate`. Both histories kept, and the **value moved 5 → 4** — the arithmetic rather than either side, since main's list is 5 long and this removes one more.

Queued as it was, this would have merged to **4 entries against a ceiling of 5** and failed after waiting for a build slot to find out.

| | |
|---|---|
| `policy-gate` probe · `shellcheck -S error` | intact · 0 |
| `mktemp -t` reintroduced? | no — the portable template survived the merge |

Entries and ceiling counted by hand as well as by the gate.

This closes the cascade of four branches that each freed one exemption from the same shrink-only ratchet — #2864, #2870 and this one all needed the same treatment after #2871 landed. Recorded as gatehouse **F-149**.

PR #2878, opened 2026-09-12
Receipts are filed under the action key they are about, and the store refuses to believe its own index

brandon-coproduct:
M1 derives an action key. M2/M3 produce a host-signed receipt. **Nothing put one under the other** — the two halves were a key with nowhere to go and a receipt nobody could find. This is that join.

## The part worth reviewing is the refusal, not the storage

A file store keyed by hex is a lookup table, and a lookup table can be wrong: a bad write, a partial copy, a directory restored from the wrong backup, a key derivation that changed underneath the stored blobs. Every one of those hands back a receipt for a **different gate** — the single failure that makes a receipt cache worse than no cache, because it is a green check citing work done on something else.

So the receipt names its own key, and `get` checks the blob it found against the key it was asked for. The filename is a hint; the content is the authority. Without that check this crate is a `HashMap` with extra steps and the design's soundness rests on a filename.

## Driven red (A-19)

Deleting the comparison in `get`:

```
tests::a_receipt_mislabelled_on_disk_is_never_served_as_a_hit --- FAILED
tests::a_receipt_that_cannot_say_what_it_is_about_is_not_a_hit --- FAILED
test result: FAILED. 5 passed; 2 failed
```

Restoring it: `7 passed; 0 failed`.

That test deliberately reaches **past `put`** and writes the wrong receipt to the right path by hand, because that is what a bad restore actually does. Checking on the way in catches a caller's mistake; checking on the way out catches everything that happens to a file after it is written, which is the larger set.

## Three distinctions the error type exists to keep

| Situation | Result | Why it must differ |
|---|---|---|
| No receipt filed | `Ok(None)` | A miss means "run the gate" |
| Blob names a different key | `Err(Mislabelled)` | Treating it as a miss re-runs forever while the corruption stays |
| Blob cannot say what it is about | `Err(Unreadable)` | A-2: "I could not look" is never "I looked and it was fine" |

## What it does not do

Verify the signature. That needs the signer's public key, which a store has no business holding. A caller that skips it has an unsigned cache and this crate cannot tell it so — **stated in the module doc rather than implied by silence.**

## Known gap, named rather than hidden

The store holds receipt JSON and requires an `action_key` field. **No producer writes that field yet.**

An earlier revision of this description said the blocker was that `pod_receipt::Receipt` lives in a binary-only crate and needed extracting. That conflated two different receipts. `pod_receipt::Receipt` is what a *pod execution* produced; the *CI* receipt is `nucleus_receipt::Receipt` carrying `Projection::Ci`, which is already a public library crate with Ed25519 `sign`/`verify` and RFC 8785 canonical bytes, and whose own doc says "the typed body lives in the consuming CI system". Nothing needs extracting. What is missing is that typed body — the plan's `nucleus-ci-verdict` — and the store learning to read the action key out of a `Projection::Ci` body rather than off the top level of an untyped blob. That is the next commit, not a refactor of `nucleus-node`.

Tests use `ActionKey::derive` with real `Inputs` rather than a byte constructor — a `from_bytes` added for tests would be a hole straight through the exhaustive destructure that makes the key total.


brandon-coproduct:
`Rustfmt` here is real and one line of work: `cargo fmt` has six diffs in this branch.

```
crates/nucleus-ci-verdict/src/lib.rs:217
crates/nucleus-receipt-store/src/lib.rs:123
crates/nucleus-receipt-store/src/lib.rs:173
crates/nucleus-receipt-store/src/lib.rs:242
crates/nucleus-receipt-store/src/lib.rs:262
crates/nucleus-receipt-store/tests/invalidation.rs:177
```

`cargo fmt --all` clears it. The `A2A example server` red is almost certainly the same cause — `RUSTFLAGS: -D warnings` is set workspace-wide, so an unformatted tree that also trips a lint fails the build there — but it is worth confirming from that job's log rather than assuming, since the two are separate jobs.

Left to you; flagging so it is not read as the shadow-gate noise on the other PRs.

PR #2892, opened 2026-09-14
rustls 0.23.45, because 0.23.43 reds audit and deny on every branch

brandon-coproduct:
RUSTSEC-2026-0285 — TLS 1.3 handshake messages incorrectly accepted across encryption level
boundaries. Affected `<0.23.45`; the tree was on 0.23.43, reached through `hickory-net`,
`reqwest`, `quinn` and `iroh`.

The advisory landed on 2026-09-14 between main's last green audit at 11:01Z and the next PR run
at 16:48Z, so it reds `audit` and `deny` on main and on every open branch at once, including
#2891 and both build-receipt drafts. Nothing merges until this moves. Dependabot's #2890 bumps
13 crates and does not touch rustls.

Lockfile only: 0.23.45 is semver-compatible, so no manifest changed. `cargo update -p rustls`
also carried `rustls-webpki` 0.103.13 -> 0.103.15.

Validation, run locally against both lockfiles rather than inferred:

- `cargo audit` on main's lock: `RUSTSEC-2026-0285`, severity 5.3 medium, "Solution: Upgrade to
  >=0.23.45", exits with `error: 1 vulnerability found!`.
- `cargo audit` on this lock: 1207 crate dependencies scanned, no vulnerability.
- `cargo deny check advisories bans licenses sources` on this lock: `advisories ok, bans ok,
  licenses ok, sources ok`.
- `cargo check --workspace --locked`: clean.

Noted, not fixed here: that `cargo deny` run also emits `advisory-not-detected` for
`deny.toml:42`'s `RUSTSEC-2026-0253` ignore. Its comment says it drops when iroh/ratatui move to
a fixed `lru`; they have — the advisory is `patched = [">= 0.18.2"]` and the tree now carries
`lru` 0.16.4 and 0.18.4, and the advisory matched no crate. That is an exemption outliving its
obstacle, and it belongs in its own change rather than folded into a security bump.


brandon-coproduct:
Superseded by #2894, which carries this exact rustls commit plus the cargo-fuzz 0.13.2 fix.

Keeping them apart did not work: this branch cannot pass `Fuzz` (cargo-fuzz 0.13.1 fails to compile on every branch), and the Fuzz fix alone cannot pass `audit`/`deny` (RUSTSEC-2026-0285). Neither goes green without the other, and gatehouse's F-147 records that stacking them is unsafe here — a stacked PR's in-flight CI dies when its parent merges, because the base branch is deleted out from under jobs that already read its name.

The rustls change is unmodified in #2894; `deps/rustls-rustsec-2026-0285` is left in place.

PR #2899, opened 2026-09-14
A validity window says when a pod runs, not what it computes

brandon-coproduct:
`program_digest` v2 already routed inline policy through the lattice's **semantic** checksum rather
than its provenance UUIDs — the right instinct, and its comment says so. But that checksum still
hashed the **validity window**, and `resolve_policy` mints a fresh window on every call. So two
invocations of one recipe produced two program identities, and every cache keyed on the digest
missed by construction.

That is what gatehouse's `result-cache-misses-4ec60ba.json` was recording: five consecutive misses
that read as "nothing stored yet" and were the key moving.

### The change

`program_checksum` is the same projection with the window omitted. `checksum` keeps it, and
certificates keep using `checksum`, **because for an authorization the interval IS the meaning** — a
signature over one interval must never speak for another. For a program identity the window is the
same category as `pod_id`, `session_id` and `cgroup`, which v2 already excludes: it says *when* a pod
may run, not *what it computes*.

Nothing is loosened. The window is still carried in the execution expectation, still checked there,
and still enforced by the runtime. Only the question "is this the same program" stops depending on
the clock.

### Tests

Both halves are pinned, and both matter:

- alike-but-for-the-window is the **same program** and a **different certificate** — dropping the
  first leaves caching impossible; dropping the second would let a signature for one interval speak
  for another
- every other field still separates two programs, so the identity has not been hollowed out to buy
  cache hits

1271 portcullis tests and 81 in `nucleus-spec` pass, strict Clippy clean.

### Measured end to end

With gatehouse pinned to this revision, two identical `build_spec` calls now produce one program
digest. That assertion **fails on the previous pin** and passes here, which is the only evidence that
matters: this is the crate whose result cache and cross-tree `ReuseKey` are keyed on it.

### A correction to the record

gatehouse#30 derived the lattice's UUID and said it addressed this. It did not: `digest_parts` never
included `id` or `created_at`, so the UUID was never in the program digest. #30's commit message is
honest that the timestamps remained, but its framing overstated the fix. The window was always the
sole cause, and this is the change that removes it.


brandon-coproduct:
Superseded by #2900 and closing.

This PR went `DIRTY` against main after #2884 landed, and its program-identity half is now carried by #2900 — cut fresh from main, so no conflict resolution is needed. #2900 also carries the two pieces this PR grew along the way: the `lattice_tests.rs` extraction that keeps the module under the sweep ceiling, and tests for `program_checksum` (three mutants survived it untested, including `==` → `!=`, which hashes only the window and inverts the fix).

The scorecard-probe blocker this PR hit is also gone: the slack probe now scales its perturbation with the family's population, and that landed on main with #2884.

Nothing here is lost; it is all in #2900 against a base that merges.

Issue #2730, opened 2026-09-09
cargo test --workspace compiles zero tests on a clean checkout: nucleus-verifier-service embeds an ungenerated wasm artifact


bcrisp:
`cargo test --workspace` and `cargo build --workspace` fail on a clean checkout:

```
error: couldn't read `crates/nucleus-verifier-service/src/../../../sdks/verifier-js/pkg/nucleus_verifier_wasm.js`: No such file or directory (os error 2)
error: couldn't read `crates/nucleus-verifier-service/src/../../../sdks/verifier-js/pkg/nucleus_verifier_wasm_bg.wasm`: No such file or directory (os error 2)
error: could not compile `nucleus-verifier-service` (lib) due to 2 previous errors
```

Nothing else in the workspace is tested when this happens: the run compiled
**zero** test binaries. 188 s of wall clock, no test suites executed.

## Cause, which CI already documents

`routes.rs:45-47` embeds the built wasm SDK:

```rust
const WASM_JS_SHIM: &str = include_str!("../../../sdks/verifier-js/pkg/nucleus_verifier_wasm.js");
const WASM_BG: &[u8]     = include_bytes!("../../../sdks/verifier-js/pkg/nucleus_verifier_wasm_bg.wasm");
```

`sdks/verifier-js/pkg` is gitignored (`sdks/verifier-js/.gitignore:2`), so a
fresh clone does not have it, and only `wasm-pack build` produces it.
`ci.yml:367-373` states this exactly:

```yaml
# wasm SDK with `include_str!`/`include_bytes!` on sdks/verifier-js/pkg/*,
# which only `wasm-pack build` produces. Without it the workspace does not [build]
- name: Install wasm-pack
  run: curl https://rustwasm.github.io/wasm-pack/installer/init.sh -sSf | sh
- run: wasm-pack build sdks/verifier-js --target web --release
```

So the requirement is known and handled in CI. It is not surfaced to anyone
building locally.

## Why this is worth fixing rather than documenting

* **Nothing tells you.** `wasm-pack` appears in `docs/split-trust.md:320` and
  `docs/verifier-service-threat-model.md:165` — neither is a place a contributor
  looks before running the tests. It is absent from `README.md`, from the
  `justfile`, and from `scripts/prepush.sh`'s own header, which advertises
  itself as *"everything CI would red in its first minute, run locally in about
  one"*.
* **The error names a path, not a cause.** A raw `include_str!` failure gives no
[...]

Issue #2557, opened 2026-09-04
[epic] Formal-methods CI shape: dedupe, cache correctness, timeouts, scope, solver selection


brandon-coproduct:
## Problem
A Lean-touching PR spends ~76 duplicated CI minutes: `lean-build.yml` builds one target that is a subset of the proven tier; `aeneas.yml`'s 13-target build is a subset except two; `research-lean-build.yml` (39.6 min median) has no timeout and hit 364.9 minutes; `coverage-matrix` re-runs llvm-cov + mutants on every merge-group entry; four workflows share one required job name. No `just`/`Makefile` target runs any proof locally.

## Decision (owner-approved 2026-09-04)
Timeouts everywhere; split the research tier (sorry/axiom grep per-PR, `lake build` nightly, 90-min cap); delete the subset workflow after re-homing its target; scope `coverage-matrix` by diff on merge_group; unique job names; per-harness Kani solver selection with `--harness-timeout`; nightly `--coverage -Z source-coverage` artifact rendered as harness line coverage, never as "% verified".

## Governing rules (apply to every sub-issue)
1. A gate that can pass without measuring is deleted or fixed before anything depends on it; every gate gets a `scripts/check-gates-can-fail.sh` probe that plants a regression and asserts red.
2. Nothing is advertised that is not run: every `#[kani::proof]` is named by a job or on a shrinking allowlist; every `lean_lib` maps to a concrete workflow step; documentation numbers are generated, never typed.
3. Two independent kernels check the proven tier; axiom sets are asserted per theorem; `implemented_by`/`extern`/`csimp` are banned in proven sources and generated output.
4. Un-extractable OS code gets typestates + fault-injection tests, never a fake Lean/Kani artifact.
5. Vendor neutrality (CLAUDE.md): external proof-search engines are named generically.

## Scope
- [x] #2568
- [x] #2569
- [x] #2597
- [x] #2586
- [ ] #2587
- [ ] #2598

PR #2910, opened 2026-09-15
fix(verifier): the wasm pin check fired where nothing was embedded

brandon-coproduct:
## What this is

`nucleus-verifier-service` could not be built from a clean checkout. `cargo build -p nucleus-verifier-service`, default feature set, panicked in `build.rs` on a missing `sdks/verifier-js/pkg/nucleus_verifier_wasm.js`.

That is the failure **#2730 removed**. It made the embed optional behind `embedded-wasm` (OFF by default) for the reason its `Cargo.toml` still records: an unconditional embed *"made a clean checkout fail to compile this crate — and one crate failing takes every other crate's test targets down with it."*

**#2897's pin check re-introduced it**, because the check is unconditional while its subject is not. With the feature off, the `include_bytes!` / `include_str!` in `routes.rs` are `#[cfg]`'d out and nothing reads those artifacts — but `build.rs` demanded them anyway, so the script was the sole cause of the failure. One line gates it on `CARGO_FEATURE_EMBEDDED_WASM`.

## Why it was easy to miss

The build script's header argued it *"adds no failure mode that did not already exist: `include_bytes!` of a missing file is already a compile error."* That is true only when the feature is **on**, which is the configuration the header silently assumed — and not the default one. The reasoning was sound about the case it considered; it just wasn't the case a contributor gets on a fresh clone. This adds both the missing qualifier and the gate.

## The product test

**More confidence**, peripheral to the frontier — no authority dimension changes, ρ and C(T) unmoved. What changes is that the repository builds for someone who has just cloned it.

## Invariants kept

**The pin check is unchanged where it has a subject.** With `--features embedded-wasm` it still refuses to build against a missing or wrong artifact — verified, not assumed (below). No default widened; nothing about what gets embedded, or what is accepted when it is, changes.

Pins guard what is embedded. With nothing embedded there is nothing to guard, and a gate that fires where its subject is absent is not stricter, only broken.

## Validation

Measured on this base, both directions:

```
cargo build -p nucleus-verifier-service            before: panics on missing .js
                                                   after:  Finished dev profile
cargo build -p ... --features embedded-wasm        still refuses (pin check intact)
cargo test -p nucleus-verifier-service             97 passed, 0 failed
cargo clippy -p ... --all-targets -- -D warnings   clean
cargo fmt --check                                  clean
```

Found while trying to verify that #2909's `CreditEvent` seal left this crate's `mint::mint_events` consumer intact — which it does, and those 97 tests are what says so. That check was impossible to run locally until this was fixed.

## Not gated, and worth saying plainly

**Nothing in CI would have caught this, and nothing added here would either.** CI builds the wasm before it builds this crate, so the default-feature configuration a contributor gets on a fresh clone is a configuration CI never assembles — the gate would need a job that builds this crate with no wasm present. That is a CI-shape change beyond a one-line fix, and #2557's remit rather than something to bolt on here. Flagging it rather than letting this PR read as if the hole is now closed: the regression is fixed, the ability to notice the next one is not.

https://claude.ai/code/session_01MUXVWHyPVkLspZcXE5gSZj

---
_Generated by [Claude Code](https://claude.ai/code/session_01MUXVWHyPVkLspZcXE5gSZj)_


brandon-coproduct:
Auto-merge armed (squash). One check is red and it is not this PR's; recording why rather than letting it merge silently past a failure.

**`The fmt gate, run the gatehouse way (shadow, informational; red = could not look, never disagreement)`** — [job log](https://github.com/coproduct-opensource/nucleus/actions/runs/35024662567/job/104568666876):

```
gatehouse: gate fmt: held          ← exit 0, verdict "held"
##[error]gatehouse: could not verify the receipt against the log:
         could not look: ProofNotAtCheckpoint { proof_size: 1247, trusted_size: 1248 }
```

The fmt gate itself **passed**: `exit_code: 0`, `verdict: "held"`, and the receipt was emitted. What failed is verifying that receipt's inclusion proof against the transparency log — a proof minted at log size 1247, checked against a trusted checkpoint at 1248. The log advanced by one entry in between. That is a race in gatehouse's log verification, in a service this diff does not touch, and it is precisely the "could not look" the check's own name reserves for this case rather than a formatting disagreement.

Three things support it not being this PR's:

- the required **`Rustfmt`** check passed on the same head, so the tree really is formatted;
- the sibling **clippy** and **ci-spec** shadow gates, same action and same run, both passed;
- the same fmt shadow gate passed on #2909, whose head is a much larger diff.

This PR changes one file — `crates/nucleus-verifier-service/build.rs` — and touches nothing that could reach a transparency-log checkpoint.

**No fix ported, because there is nothing to port:** the failure is a timing race in the log service, not a defect with a change behind it. I have not re-run the job — a re-run would very likely go green simply because the checkpoint has since advanced, which would tidy the badge without establishing anything. If someone wants the record clean, re-running is harmless; the outcome just shouldn't be read as evidence either way.

`mergeable_state` is `unstable`, not `blocked`: every required check passes, and this shadow gate is informational by design. Auto-merge will fire on the required set.

---

PR #2926, opened 2026-09-16
fix(perf): podburst, toolcall and agency reach the node over mTLS

brandon-coproduct:
fix(perf): podburst, toolcall and agency reach the node over mTLS

All three signed node requests with the HMAC auth secret, which the node stopped
accepting in Move B ("mTLS with SPIFFE is the only authentication method left",
`nucleus-node/src/auth.rs`). Each failed at its first node request against any
current node, and no workflow runs them, so nothing noticed. Found while
building `guest-transcript` (#2923), which had to speak mTLS from the start.

`node_mtls` is now the harness's one node client: a flattened
`--tls-cert/--tls-key/--trust-bundle` group (env `NUCLEUS_NODE_TLS_*`, the same
names `nucleus node` uses) and create / list / cancel over reqwest with the node's
CA as the only root. A partial identity is refused by name before connecting.
`podburst`, `toolcall`, `agency` and `guest-transcript` all use it; the HMAC node
helpers, `--auth-secret` on those commands, and `podburst`'s signing-only
`--actor` are gone. Node URLs default to `https://`. Calls to a pod's tool-proxy
are unchanged: the node's signed proxy fronts them, and `agency --local` signs
its own.

Live, nucleus-kvm: `podburst --counts 1,2` authenticated at preflight and booted
clean bursts of 1 and 2; `toolcall` created a pod, ran its checks through the
proxy ("all checks passed") and cancelled it; with no identity the command stops
with the named refusal.

Stacked on #2923 (it edits the same `main.rs` and refactors
`guest_transcript`'s client into this module).

**Review note:** stacked on #2923; the first three commits are that PR's. Once it merges, this diff is the last commit only.


brandon-coproduct:

PR #2932, opened 2026-09-16
test(perf): cancel is a barrier on a live node (held with #2930, violated without)

brandon-coproduct:
Stacked on #2926 (mTLS harness client), which is stacked on #2923. It measures the fix in #2930 and needs no node change of its own.

**What:** `nucleus-perf teardown-barrier`, plus the guest mode `nucleus-adversary-probe barrier`. This is the live half of #2930's in-process census, which had no VMM, broker or network teardown inside the window.

**How:**
- The guest holds a workload-API connection and sends `FETCH_SVID` as fast as it is answered until it is killed.
- Once the guest is mid-stream, the harness cancels the pod.
- It then counts the commands the **node's own log** records for the pod after the line where the pod's bridge began shutting down.
  - The count uses line order, not timestamps across two clocks, and never takes the guest's word for anything.
  - The allowance is 2: one in-flight frame per possible connection.
  - Lines are matched by the JSON log's `fields.message`. Matching raw lines finds nothing and would read NOT CHECKED, which a unit test pins.

**Live on nucleus-kvm:**

| Node | Runs | Commands before shutdown | After |
|---|---|---|---|
| with #2930 | held 4/4 | 467–917 | **0** |
| bridge as before #2930 | **VIOLATED 3/3** | 286–579 | **8, 12, 17** |

Exit status: 0 held, 1 violated, 2 could not look.

**Also found, not fixed here:** the pod reaper re-reaps an exited pod every 10 s while it stays registered.
- Each pass appends another `pod_exited` entry to the lifecycle log; one exit was recorded 6 times.
- Each pass logs "registration and removal keys have drifted apart", because the identity was already released. The message is misleading.
- `release_child` is idempotent, so budgets are not double-released.


brandon-coproduct:

PR #2936, opened 2026-09-16
feat(jsonl): an acknowledgement is built from a durable append's proof

brandon-coproduct:
Stacked on #2935.

**What:** `nucleus_jsonl` now exposes `append_line_synced -> Durable` and `append_line_unsynced -> ()`, plus async forms, in place of a `Durability` argument. `Durable` has a private field and is `#[must_use]`, `!Clone`, `!Copy` (ADR 0007 C-1/C-4). The two acknowledgements a collector sends take it **by value**:
- SHIP_RECEIPT's `{"status":"collected"}`, via `receipt_collected(Durable)`;
- the Article 12 collector's `204`, via `record_accepted(Durable)`.

So acknowledging a record that was never appended, or only reached the page cache, doesn't compile.

**A-19**, each probe applied to the real handler:
| Probe | Result |
|---|---|
| acknowledged receipt appended without sync | `E0308` |
| acknowledgement from a forged proof | `E0451` (private field) |
| one proof acknowledging twice | `E0382` (moved) |

There's also a `compile_fail` doctest for forging outside the crate. Workspace clippy is clean and the collector tests pass.

**Limit:** a separate literal reply string elsewhere is still writable. The witness makes the honest path the only one that type-checks at the two existing sites; it doesn't forbid new ones.


brandon-coproduct:
Update: CI's scorecard (`life` family) flagged `Durable` as an unbounded affine right. That was a real hole: the proof didn't name which record it proved, so one record's proof could acknowledge another. `Durable` now carries a `state_hash` of log + record, and both acknowledgement sites check it (`Durable::proves`). A-19: with `proves()` ignoring the record, both site tests go red. `life` is 6/12, pinned.

PR #2940, opened 2026-09-16
fix(node): one teardown per driver; exited local pods stop their proxy, cancelled containers keep exit state

brandon-coproduct:
T2 of the effects/walks plan.

**Duplication:** each pod driver had two teardowns, `cancel` and `cleanup` (on exit); Firecracker's shared 24 of 27 lines. The copies drifted, and two bugs sat in the gaps:

| Driver | Bug |
|---|---|
| Local | `cleanup` did nothing while `cancel` shut down the signed proxy, so **a pod that exited on its own kept its proxy listening** for the node's lifetime |
| Container | `cancel` removed the container **without caching its exit state** (`cleanup` did), so a cancelled pod's `status()` read `Error`, and the reaper audited its exit as "No such container" |

**Fix:** one `teardown(stop)` per driver. `Stop::{Kill, AlreadyExited}` is the only difference between the two paths; `PodHandle::cancel` and `cleanup_after_exit` both call it. Firecracker's teardown order is written once. The container caches its exit state before removal on both paths.

**Test:** a local pod whose process exits on its own, holding a real `SignedProxy`, must refuse connections on the proxy port after `cleanup_after_exit` (non-vacuity: it accepts before).

**A-19:** red on main ("the signed proxy of a pod that exited on its own is still listening"), green here.

**Not driven red:** the container fix. There's no Docker locally and no container-driver tests in the repo. It is correct by construction but unmeasured.

**Lines:** `main.rs` 3748 → 3721 (−27 production); +55 test. nucleus-node tests (537) pass; clippy clean under default and `local-driver` features.


brandon-coproduct:
The container fix is now driven red/green against a real Docker daemon (Docker 29.1.3 in the nucleus-kvm Lima VM), via a new `#[ignore]`d test, `a_cancelled_container_reports_its_exit_not_an_error`. It starts a real container, cancels the pod, and requires `status()` to be `Exited`.

- **With this PR:** passes.
- **`main.rs` as on main:** fails with `Error { "Docker responded with status code 404: No such container: 6eb675d3…" }`, the exact defect.

Run it with `NUCLEUS_TEST_DOCKER_IMAGE=<local image> cargo test -p nucleus-node --bin nucleus-node -- --ignored a_cancelled_container`. It's ignored by default because CI has no Docker daemon, so CI doesn't run it; that remains a gap.

PR #2942, opened 2026-09-16
test(perf): nucleus-perf stress — concurrent tool calls checked for linearizability

brandon-coproduct:
Stacked on #2932 (→ #2926, #2923). S0 + S1 of the stress plan: the tool-proxy stressed as an IFC-enforcing web server, **with an oracle**.

**What:** `nucleus-perf stress` spawns a real `nucleus-tool-proxy` under a compiled, sealed grant, reusing `agency::spawn_local_under_grant` (not copying it). `--clients` threads send signed calls from a closed vocabulary (read/write two files, a blocked path, glob, run). Each call records its invoke and return instants and a normalised result. The history must be **linearizable** against a sequential model.

**Checker** (`stress/linearize.rs`): Wing–Gong search with Lowe/Porcupine memoisation, plus:
- Porcupine-style **partitioning** (per-file registers). Four clients leave no quiescent point, so without it a run is one window.
- **Quiescent windows** carrying the set of reachable states.
- An arbitrary-size done-set.
- A **state budget** whose exhaustion is "could not look", never a pass.

Unit tests on a one-shot approval model: double spend rejected; refused concurrent spend accepted in either order; non-overlap forces order; a 300-call non-quiescent window is searched.

**Calibration** (`--clients 1`) and larger runs found three fail-closed refusals, with no effect, that the model must allow: 402 `budget_exhausted`, 413 (8 MB values), 429.

**Measured (unmodified proxy):** linearizable at 1/4/8/16 clients, up to 960 calls and 1 MB values, across seeds.

**A-19:** a torn write injected into `Sandbox::write_internal` (half, 50 ms, rest) is caught: reads returned `"val"`, **NOT LINEARIZABLE**.

**Limit, stated:** with a 3 ms window the probe stayed green, because no read landed in it. History checking finds a race only when the schedule hits it. The proxy's real write (truncate then write) is non-atomic with a microsecond window the harness hasn't hit, even at 1 MB.

**Not yet covered:** IFC taint. Under a compiled grant no local taint source is reachable: web effects are host-scoped to real registries, and command output doesn't gate writes. That needs a spec-based launcher with a DNS allowlist for a local mock (next PR), and so do the plan's race targets (approval peek-then-spend, taint ingested after I/O).

Exit status: 0 held / 1 violated / 2 could not look.


brandon-coproduct:
**Update: `--policy trifecta` (S1b).** IFC is now under concurrent stress: a local mock web server serves a prompt injection, and `research-web` + write/edit puts all three trifecta legs in one session.

**The model, as measured** (calibrated sequentially, then every concurrent failure run down to its cause from a `--history` dump):
- **Taint takes effect at delivery, not decision.** Concurrent fetches decided before either delivers are both admitted and both reach the network.
- **Exposure precedes taint.** Concurrently, a write between a fetch's admission and its delivery gets `approval_required`. No sequential test can see this state.
- **The breaker's lockdown is sticky.**

**That last law was violated on main:** lockdown lifted itself ~400 ms later. It's fixed in **#2943** (security): 3/40 not linearizable on main → 0/40 with the fix.

Also noted: the lockdown refusal is reported as `400 body_error`.

brandon-coproduct:

PR #2957, opened 2026-09-17
gatehouse: the plan declares the machine, and the machine is held to it

brandon-coproduct:
**Draft: blocked on gatehouse #46 + #49 merging and `GATEHOUSE_REF` moving with them.** The plan imports the prelude by content digest, and `xtask gatehouse-pin` checks that `GATEHOUSE_REF` and the import digest are one fact, so this cannot go green until the gatehouse side lands. Stacked on #2956 (the Lean bump), because `lean-build` declares `lake@4.30.0` / `lean@4.30.0`.

## What this is

The pipeline said what each gate runs and nothing about what it runs **on**. Five gates, one Mac VM, and every capacity question about that pairing lived in an operator's head. gatehouse's prelude now decides those questions (`platformMatches_b`, `memoryFits_b`, `builderFits_b`, `timeoutCoherent_b`, `toolsCovered_b`, `diskFits_b`, `builderBudget_b`, `seedPinned_b`, `scopeClosed_b`), so this plan declares the operands.

Every number is measured on the builder (2026-09-17) or read out of a file that is itself pinned:

| field | value | where it comes from |
|---|---|---|
| `platform` | `linux/arm64` | the lane is a Mac VM. Four gates were moved off `linux/amd64` **by hand** in #2952; `lean-build` was left behind |
| `memMb` | 4096–8192 | `lean-build` asked for 16384 on a lane that gives 8192 — the executor **clamps** rather than refusing. Measured at 8192: 2m25s |
| `diskMb` | 4096–16384 | largest gate scratch measured was clippy's, 2.0 GiB |
| host | 78284 MiB disk, 23981 MiB memory, 2 lanes | `df -m /`, `/proc/meminfo`, the two worker units |
| catalog | keep 3 × 4096 MiB, 10240 MiB headroom | the worker's retention policy and a measured prepared input (4058 MiB) |
| queue | 21600000 ms, 2 in flight | the 360-minute timeout in `ci/merge-queue.toml`; `max_entries_to_build = 1` plus the head under test |

**The queue capacity is exactly at its limit, and saying so is the point:** 10800000 ms of gates × 2 commits = 21600000, and one builder × the window = 21600000. `builderBudget_b` holds today and refuses the next gate that lands on this builder — the true statement about a one-machine lane.

## What is declared empty, and why that is not "nothing"

- `seeds`: no gate is restored from prepared scratch yet. `lean-build` will need a digest-pinned Mathlib olean cache before it can run in a microVM with no network.
- `reads`: the scan resolving nucleus's **108** `include_str!`/`include_bytes!` sites is not written, so `scopeClosed_b` holds vacuously here — the clause is in place and its operand is not, exactly as `tools` was for one release. Three of those sites cost a full gate run each to find this week.

## Gate definitions

Two corrections to the files that actually run: `lean-build`'s platform (`linux/amd64` on an arm64 lane) and its memory (16384 → 8192, what it was going to be given anyway), plus `disk_mb` brought in line with the plan. `resources` is excluded from a gate's hash and `env` is not, **so `lean-build`'s gate identity changes and the other four gates' do not** — no already-decided commit loses its verdict.

`tools` is not yet in the JSON: `GateDef` gains the field in gatehouse #47, and adding it before that ships would be rejected by the deployed controld.

## Merge order

1. gatehouse #46 → #49 (one unit: #46 is red alone on the cost ratchet, #49 fixes it)
2. nucleus #2956 (Lean v4.30.0)
3. bump `GATEHOUSE_REF` in `gatehouse-plan.yml` and `gatehouse-shadow.yml` to the merged gatehouse commit, in this PR
4. `PUT /v1/nucleus/plan` with this plan and these gate definitions


brandon-coproduct:

**Its gatehouse content is already on `main`** — the policy declaring the machine (platform, hostDiskMb 197316, hostMemMb 23981, lanes) landed by another route, so there is nothing here to re-land.

**It cannot be merged as it stands.** `plan/declare-the-machine` is 6 commits behind `main`, and `git diff main plan/declare-the-machine` shows it would *delete* work that landed afterwards (`crates/nucleus-tool-proxy/src/web_fetch_policy.rs`, among others). Merging it would be a revert wearing a feature's clothes.

Nothing is deleted by closing — the branch `plan/declare-the-machine` remains, and this can be reopened if I have misread what it carries.

PR #2958, opened 2026-09-17
gatehouse: the eleven text gates move to the lane, and CI-RP follows a relay

brandon-coproduct:
**Draft, stacked on #2957** (same gatehouse-merge dependency). Wave two of the cutover.

## The wave

Eleven `scripts/check-*.sh` text gates — nine of them required contexts — become **one gate** that runs the same eleven scripts. Its exit is their conjunction, and each script runs even after an earlier one fails, because CI gives every step `continue-on-error: true` and reports eleven contexts separately: a gate that stopped at the first failure would decide less than the eleven it replaces.

**Measured in the gate image on the builder, offline, over a real checkout: all eleven pass in 5 seconds.** The declared timeout is 120 s — twenty-four times the work, because the floor here is a microVM's startup, not the greps. Queue budget goes to 5520000 ms of 21600000.

## CI-RP would have passed all eleven vacuously

Each of these contexts is produced by a **relay** job whose only command is

```bash
case "$STEP" in success|skipped) exit 0;; *) echo "::error::…"; exit 1;; esac
```

with `STEP: ${{ needs.text-gates.outputs.verify_strict }}`. Comparing a replacement against *that* finds nothing to cover. So CI-RP now follows a producer that decides nothing of its own: `needs.<job>.outputs.<n>` → that job's `outputs:` → `steps.<id>.outcome` → the step's `run:`. An unresolvable relay is **CI-RP-4**, never a pass.

Two parsing defects surfaced while doing it, both of which had made relays invisible: a `case` arm's label read as a program (`success)`, `skipped)`), and `sh -c "…"` — the shape of nearly every gate's argv — read as running `sh`.

## One real defect, found by running the gates where they will run

`check-north-star-ledger.sh` terminates its table scan with `/^#{2,4} /`. Interval expressions are a POSIX awk feature **mawk does not implement**, Debian's awk *is* mawk, and under it the terminator never matched: the script read the whole document instead of one table and reported **18 defects in a clean tree** — while passing on a GitHub runner, which has gawk. Fixed portably (`^###?#? `), and gatehouse#50 records awk's implementation in the tool inventory so the next such difference is a refusal instead of a puzzle.

## Verification

- `cargo xtask ci-spec check`: clean with all eleven replacements declared.
- Drop one script from the gate → **exactly that context** reds with CI-RP-2.
- `gate plan check`: six gates, admissible under all fifteen clauses.
- 11/11 scripts green in `gatehouse/nucleus-gate-tools:inventory-rg-jq`, `--network none`.

Not included: the `ci/required-checks.txt` swap and the ruleset edit. Those are the owner's step, and they come after this and gatehouse#50 are deployed.


brandon-coproduct:

**Its gatehouse content is already on `main`** — the `text-gates` gate itself landed by another route, so there is nothing here to re-land.

**It cannot be merged as it stands.** `plan/wave-text-gates` is 6 commits behind `main`, and `git diff main plan/wave-text-gates` shows it would *delete* work that landed afterwards (`crates/nucleus-tool-proxy/src/web_fetch_policy.rs`, among others). Merging it would be a revert wearing a feature's clothes.

Nothing is deleted by closing — the branch `plan/wave-text-gates` remains, and this can be reopened if I have misread what it carries.

PR #2964, opened 2026-09-18
perf(gatehouse): the three compiling gates are restored from the pinned seed

brandon-coproduct:
The warm scratch has existed since 10:24 today — 8 GiB, digest-pinned, at `/srv/gatehouse/seeds/warm.ext4`. The guest mounts it read-only through an overlay (#2961, merged 14:35), the controller restores from it, and **not one gate named it**: zero of the last twenty gate recipes carried a seed, so every run compiled from nothing.

Measured on the lane today:

| gate | wall |
|---|---|
| test-core | 686–728 s |
| ci-spec | 286–320 s |
| clippy | 285–289 s |
| fmt | 34–76 s |
| text-gates | 36–40 s |

The three that compile are ~1300 s of the lane's ~1450 s of work, and the lane is the whole critical path of a merge — GitHub's other 88 checks all finish inside 9 minutes in parallel.

So `clippy`, `ci-spec` and `test-core` name `warm@sha256:f84b9b68…` and the policy lists it, which is what `ci.seedPinned_b` requires: an operator reviews the list, a gate may name only what is on it, and the digest is checked against the bytes before a VM boots.

**It re-identifies nothing.** `gate expect` gives test-core the same `gate_def` `aae31189f8290a7eb6133029581dd023256774f386f5d0553cd0c237c2c265ef` with the seed and without, because `GateSemantics` excludes seeds deliberately — a cache changes how long a gate takes, not what it concludes, and hashing it would void every receipt on every open pull request each time the cache is refreshed. The honest binding is stronger and elsewhere: the seed's digest lands in the pod's `data_digest`, which **is** inside the program identity the execution signs.

The plan's prose said seeds were empty "and that is not the same as nothing". It is no longer empty, so it no longer says that.

Based on #2958 because that branch's plan is the one the lane is actually running (byte-for-byte identical to what `controld` serves).


brandon-coproduct:
Marked draft until the executor can honour this.

Declaring a seed here is only half of it: the controller has accepted `--seed-store` since seeds existed and **the agent never passed it**, so a gate naming `warm@sha256:…` would have its name written into the recipe and still compile from nothing. This PR would have put a claim into the live plan that nothing implements.

gatehouse `feat/seed-store-wiring` fixes that end — and makes the ordering strict, because with it a gate whose plan names a seed on a worker with no `seed_store` is **refused** in `admit` rather than silently compiling cold. So merging this first would not be inert, it would red the lane.

Order:
1. merge gatehouse `feat/seed-store-wiring`, rebuild + restart the agents
2. add `seed_store: /srv/gatehouse/seeds` to both worker configs
3. then this, and the plan PUT

Ready to un-draft once 1 and 2 are done.

brandon-coproduct:

**Its gatehouse content is already on `main`** — the pinned warm seed on the three compiling gates (`warm@sha256:f84b9b68…`) landed by another route, so there is nothing here to re-land.

**It cannot be merged as it stands.** `perf/seed-and-split` is 6 commits behind `main`, and `git diff main perf/seed-and-split` shows it would *delete* work that landed afterwards (`crates/nucleus-tool-proxy/src/web_fetch_policy.rs`, among others). Merging it would be a revert wearing a feature's clothes.

Nothing is deleted by closing — the branch `perf/seed-and-split` remains, and this can be reopened if I have misread what it carries.

PR #2971, opened 2026-09-19
gatehouse: text-gates reads a seventh of the tree, and says so

brandon-coproduct:
Stacked on #2969 (base `ci/replacement-parity`) — it touches the same plan files. Retarget to `main` once #2969 lands.

Two measured speed changes. Numbers from the lane on 2026-09-19.

## Where the time goes

Required-gate critical path is ~1934 s. Per-step timing from `run/step-*/timing.json` on a real text-gates attempt:

| component | per step | note |
|---|---:|---|
| microVM lifecycle | ~20 s | boot → verified artifact, for a `grep` |
| preparation | ~5 s | re-hashes the 8.6 GB rootfs each step |
| checkpoint | ~5 s | |
| clone **verify** hash | 4.33 s | 8 GiB logical, sparse |
| seal re-hash + rest | ~15 s | |

Eleven scripts doing wildly different work all cost ~50 s — that uniformity is the tell that it is fixed per-step overhead, not the script. The clone itself is nearly free (`copy_seconds` 0.044, only 113 MB allocated), so this is the cost of **eleven pods**, not eleven forks. Making hashing free would buy 9 s of 50 s; that change is not in this PR.

## 1. text-gates declared `**` and does not read `**`

A verdict is keyed by the hash of the content in the gate's scope, so `**` means every tree is a new key and the gate can **never** reuse a verdict — on any commit, for any reason. At 713 s it is ~37% of the critical path.

The eleven `scripts/check-*.sh` were read: every `rg`/`grep` is path-restricted, and the three taking a `SCOPE_DIRS` array expand it to `crates/*/src`. The union of what they open:

```
.github/workflows/**  crates/**  docs/**  scripts/**  tests/**
kani-divergence.toml  sandbox-trusted-base.txt
```

The last two are top-level manifests opened by name — exactly the read the plan's own comment says costs a full gate run to discover.

**Verified by construction, not by reading.** A tree containing only the narrowed scope was built from this commit and all eleven scripts run against it: **11/11 pass**. (First attempt showed 10/11 — that was my harness: `git init` with no index, so the script's `git ls-files` found nothing. Populated the index, 11/11.)

Worth stating: a too-narrow scope here fails **closed**. The pod's disk *is* the scope selection, so a file outside it is absent and the script reds. This cannot produce a false green.

## The ceiling had to be widened to allow a narrower gate

Not the direction you expect. `leqCap` orders pattern sets by `subsetGlob`, and `prim_glob_subsumes` is **syntactic**:

```
scripts/**  ⊆ **  = false
crates/**   ⊆ **  = false
**          ⊆ **  = true
```

So the ceiling must list verbatim every pattern a gate names. Declaring `**` was free because `**` was in the ceiling; declaring the seven real paths was refused by `bounded_b` until `docs/**`, `kani-divergence.toml` and `sandbox-trusted-base.txt` joined it. **The lattice charges for precision.**

This is not a weakening — the ceiling already carried `**`, the top of that coordinate. The additions only let a gate say something narrower than everything.

## 2. Two long gates were using a third of the host

test-core and clippy each asked for 2 CPUs on a 6-CPU host with `inFlight` 2 — four of six cores idle while the two longest gates (702 s, 304 s) ran. Three each packs the box exactly. `resources` is outside the gate identity, so this **preserves every existing verdict**.

## What I did not do, and why

**test-core keeps `**`, deliberately.** I claimed earlier it doesn't read `.github/` — that was wrong. It runs `cargo nextest --all-features` over the whole workspace, and the workspace contains `crates/ci-spec`, whose tests read `.github/workflows/**` and `ci/**`. Proving which of forty crates' tests touch the live tree is a bigger job than it looked, and the failure mode is a false green, so it is recorded rather than guessed at.

## Verification

- `gate plan check` at pinned ref `067869b5`: `admissible: proved by admissible`.
- Only text-gates re-identifies (`f44427df` → `43d95f2c`); the other five hashes are byte-identical to the pristine elaboration.
- `.gatehouse/plan-gates.json` regenerated with `gate plan gates`, not hand-edited.
- `scripts/check-gate-defs-match-plan.sh`: `OK: 6 gate(s) carry the cmd, tools, seeds, outputs and scope the plan declares`.
- `scripts/check-ci-spec.sh`: `ok: every invariant holds`.


brandon-coproduct:
Two honest caveats on the CPU change (#2), so they are on the record rather than in my head:

**The gain is projected, not measured.** test-core at 702 s was measured at 2 CPUs. I have not measured it at 3. `cargo nextest` parallelises well, so ~1.3–1.4× is the expectation, but the only way to know is to run it. Worth checking against the first receipt after this lands.

**It is mild oversubscription.** Admission on this host is by huge-page budget, not the whole-host mutex (`pages_for(memory_mib)` against a pool of 8192 × 2 MiB = 16 GiB), so the concurrency ceiling is set by memory, not by `inFlight`:

| combination | pages | fits 8192? |
|---|---:|---|
| clippy + test-core | 4096 + 4096 | exactly |
| test-core + fmt + ci-spec | 4096 + 2048 + 2048 | exactly |
| test-core + fmt + text-gates | 4096 + 2048 + 1024 | yes, with slack |

So three small gates can co-run. At 3 CPUs each that is 9 vCPUs on 6 cores. Note the host is *already* oversubscribed under the current setting — three concurrent gates at 2 CPUs is 6 on 6, and four is 8 on 6 — so this widens an existing condition rather than introducing one. Firecracker vCPUs are host threads and the scheduler handles it, but it will add variance.

If that variance shows up, the narrower version of this change is to raise only test-core (the 702 s critical-path gate) and leave clippy at 2.

brandon-coproduct:
**Correction: the CPU change in this PR is currently a no-op.**

The worker config carries `gate_max_cpu_cores = 2`, and the agent does

```rust
let cores = spec.gate.resources.cpus.clamp(1, u32::from(config.gate_max_cpu_cores));
```

so `resources.cpus: 3` is clamped straight back to 2. Raising it in the plan does nothing until that config is raised on the builder. The change is still correct and worth keeping — it just needs the host-side ceiling to move with it.

---

**More importantly, the premise behind it was wrong.** I measured clippy's three steps properly:

| step | pod (VM execution) | total |
|---|---:|---:|
| step-0 | 65 / 83 / 92 s | 87 / 161 / 521 s |
| step-1 (workspace clippy) | **241 / 275 / 241 s** | 974 / 535 / 342 s |
| step-2 (portcullis) | 65 / 65 / 83 s | 193 / 148 / 368 s |

Clippy itself is stable at ~240 s. All the variance is outside the VM. Accounting for one attempt:

```
total 973.5 − pod 241 − prep 23.7 − ckpt 10.5 − clone 5.9 = 692 s unaccounted
```

That 692 s is **admission queueing**. `nucleus-gates.json` has `page_budget = None` and `launch_lock = /srv/gatehouse/launch.lock`, so admission is the whole-host mutex — one pod of any size at a time — and the controller's own comment says it is "needlessly strict... `lanes = 2` describes a concurrency the executor never provides". Observed directly: three processes queued on `launch.lock` behind one running pod.

**A gate takes that mutex once per step**, so clippy queues 3 times and text-gates 11 times. Roughly 70% of clippy's 1253 s is scheduling, not compilation.

Consequence for this PR: giving a gate more CPU does not help when it is waiting for a slot rather than for cores. The ordering should be counted admission first, then re-measure, then size CPUs and timeouts against the numbers that regime produces — measuring now would size everything for contention that is about to disappear.

PR #2975, opened 2026-09-20
Measure each artifact once, and name the 63% of pod boot nothing measured

brandon-coproduct:
Measured over **843 pod boots** on the acceptance builder, from the node's own `boot_trace` events:

| | |
|---|---|
| wall | **21.4 s** median, 35.7 s p90 |
| `attestation.hash` | 5.5 s, opening at **+13.4 s** |
| `proxy.health_wait` | 2.1 s |
| every other stage | under 50 ms |
| **unaccounted** | **13.4 s — 63% of the boot** |

The unaccounted time wasn't spread thin. It was **one contiguous gap**: `prepare_jail` closed at +82 ms and `attestation.hash` opened at +13373 ms with nothing instrumented between them.

What lives there is `image_identity::verify`, which calls `measure_artifact` for **every pinned artifact — kernel, rootfs, scratch, data**. The rootfs is **8.0 GiB**. `compute_attestation` then hashes the kernel and rootfs *again*, and that's the 5.5 s that was visible.

> A pod spent roughly 19 of its 21.4 seconds computing SHA-256 over 8 GiB files, twice — while the guest itself booted in **1.97 s**.

## Two changes

**`image.verify` gets a boot stage.** `boot_trace.rs` says a large unaccounted *"is a finding (unmeasured work), not an error to hide — a profiler that always sums to 100% has stopped being able to say it missed something."* It was right; this is the gap it was built to surface.

**`measure_artifact` memoizes on file identity** — device, inode, size, mtime to the nanosecond. That's the same discriminator #2972 gave the whole attestation, and the same one gatehouse's catalog uses for its verification stamps. This **extends an accepted trust to the call site still paying full price**, rather than introducing a new one. Both call sites share it, so even the first pod reads each file once; the memo is per-process, which is what makes "verify once per boot" true without a stamp file.

## What it does not survive

A writer that restores size, inode and mtime around a modification. That exposure is identical to the one the attestation cache already accepts at identical granularity — and it's now pinned by a **test** rather than only described.

`an_unchanged_file_is_measured_once` rewrites the contents while restoring size and mtime, and requires the **stale** digest back, which no amount of re-reading could produce. Asserting that two measurements of an untouched file agree would pass with no cache at all — the shape of a test that cannot fail. Verified by perturbation: disabling the memo lookup reds it.

## Why it compounds

Every gate pays pod startup. `text-gates` alone boots eleven pods — roughly 240 s of its 713 s measured runtime is pod boot.

`just prepush`: 18/18.


brandon-coproduct:
Parking this: **the memo defeats a security property, and a pre-existing test catches it.**

`Workspace build + tests` fails on `tests::admit_posture_one_byte_of_drift_reds_the_gate` (`crates/nucleus-node/src/tests_main.rs:670`). That test writes `artifact v1`, mints a posture claim for its digest, rewrites the file as `artifact v2`, and requires admission to **fail**. Its comment states the stake plainly:

> the property that makes the gate bind to the artifact, not to the pod's word.

Both strings are 11 bytes, `tokio::fs::write` reuses the inode, and the path is unchanged — so `dev:ino:size:mtime` is identical wherever mtime granularity is coarser than the gap between the two writes. The memo returns the first digest and a changed artifact is admitted.

This PR's own comment anticipated the shape and then drew the wrong conclusion:

> WHAT THIS DOES NOT SURVIVE... a writer that restores size, inode and mtime around a modification defeats it. That is the identical exposure the attestation cache already accepts at the identical granularity, **so this adds no new trust**.

The last clause is false. The attestation cache accepting that exposure did not make `admit_posture` accept it — `admit_posture` was measuring every time, and this makes it stop. The exposure is not extended, it is **introduced at a call site that did not have it**.

Worse, `measure_memo_tests` asserts the stale digest is the *correct* answer, so this PR ships a test that pins the unsound behaviour as intended.

The optimisation is still worth having — 63% of a 21.4 s pod boot is this function, hashing the same 8 GiB rootfs four times in one `image_identity::verify`. But the memo has to be scoped to that one call, where the file provably cannot change between reads, rather than installed globally inside `measure_artifact` where admission also lives.

Reworking rather than merging.

brandon-coproduct:
Superseded by #3009.

This draft chased the 63%-of-pod-boot hashing cost with a memo of its own, keyed on `dev:ino:size:mtime`. I parked it because the key defeated `admit_posture_one_byte_of_drift_reds_the_gate` — two 11-byte strings sharing an inode.

That was the right instinct for the wrong reason. The key is not merely awkward for one test; it is **unsound**, and gatehouse F-161 is the observed counterexample: a file read back as NUL bytes with device, inode, size and mtime all unchanged, because ext4 committed the inode and lost the data blocks. `git status` called the tree clean for the same reason.

Looking for other places that treat stat as a content proxy turned up #2972, already merged, using the same key for the launch attestation — which goes into the pod's certificate and is what a relying party reads to learn which kernel and rootfs ran.

#3009 removes that one and buys the speed the sound way: `image_identity::verify` already reads every pinned artifact in the jail after placement, so it now returns what it read and the attestation reports those bytes instead of hashing the same files a second time. No cache, one read, and the certificate reports the bytes that were held to the pin.

PR #2976, opened 2026-09-20
ADR 0009: the public/private line, and a gate that decides it

brandon-coproduct:
## What

Writes down the boundary between this repository and the private repositories built on it, and gives nucleus a gate that decides it from its own side.

The rule has been operating for months, enforced from exactly one direction: a private sibling runs a script that fails if one of its crates reaches nucleus. Nucleus runs nothing. The property holds here by habit — and a single private `git =` dependency would compile, test, red no existing gate, and be discovered by a stranger who cannot clone the repo.

## The rule

Nucleus holds everything a stranger needs to **check** an agent's authority and the mechanism that decides it. Everything required to operate a business on top lives elsewhere. A component belongs here iff it is **neutral** (no vendor, tenant, price list or operator identity), **key-free and fund-free** (verifying a signature is in scope; being the production signer is not), and **checkable** (its output is a proof, a receipt, or a decision recomputable from declared inputs).

The dependency arrow is one-way — the monotonicity condition on `Vis : Crate → {public ⊑ private}`. And the right to check is never sold.

The ADR states the rule and deliberately names no private repository. The inventory is recorded privately: the rule is a commitment we make, the map is a fact about a portfolio.

## The gate

`cargo xtask visibility`, and it needs both halves:

- **A manifest sweep cannot see everything.** The tempting argument is that crates.io forbids published packages from declaring `git`/`path` deps, so a non-registry source could only be declared here. That holds for registry packages and fails for git ones — `dlc-d` pulls `dlc-d-macro` from the same repository and no manifest here names it. The gate found this on its own first end-to-end run, refuting an argument the module doc had just made; the doc now records the correction.
- **A root-level `cargo metadata` cannot see everything either.** This repository carries thirteen satellite workspaces.

Population is asked of `git ls-files`, not of a directory walk: the walk found 25 workspace roots where there are 13, because a nested `git worktree` in the working copy reads as repository content. A gate whose population depends on what is lying around gives different verdicts on two clones of the same commit.

Git dependencies are permitted only where **enumerated one by one**, never by pattern — a repository that is public today can be made private tomorrow with nothing here changing. Five names across two upstreams: the four `dlc-*` crates from public `delegation_calc`, and `clippy_utils`, which the dylint passes link against and which is deliberately never published.

## A-19

Driven red on the real defect before the green was trusted: a `coproduct-private` git dependency added to a satellite's `[dependencies]` produced `FAIL … declares spiffy-hub: git dependency on …` and a non-zero exit; removing it restored the green line. The first placement of that probe landed in `[package.metadata.rust-analyzer]` and was correctly *not* flagged — which is why `every_dependency_table_is_read` exists.

## Verification

- `cargo test -p xtask visibility::` — 12 pass, including non-vacuity on both the manifest population and the satellite set (the declared list must equal the tree exactly, both directions).
- `cargo run -p xtask -- visibility` — `ok: 948 package(s) resolved, all crates.io or in-tree; 107 manifest(s) across 14 workspace(s)`.
- `cargo fmt --check`, `cargo clippy -p xtask --all-targets` — clean.

Also fixes README's link to `nucleus-code`, which pointed at a `coproduct-opensource` URL that does not exist while the same sentence called the repository private.

https://claude.ai/code/session_01D5V6CuhLqCGCaLXwPzhtf8


brandon-coproduct:
Renumbered to **0009**: [#2973](https://github.com/coproduct-opensource/nucleus/pull/2973) opened three hours earlier and claimed 0008. Two files named `docs/adr/0008-*.md` would not have conflicted in git — the names differ — so the collision would have survived a merge and been discovered by a reader holding two ADRs with the same number.

PR #2980, opened 2026-09-20
Text scans are not worth a microVM

brandon-coproduct:
**Stacked on #2974.**

`text-gates` ran eleven steps in the gatehouse lane — **eleven Firecracker pods, each paying ~22 s of boot** — to execute grep and awk over declared globs.

| | |
|---|---|
| measured in the lane | **713 s** per run |
| over 399 merges | **79 of the lane's 317 gate-hours** |
| cacheable | 5.8% |
| the same scripts, in `ci.yml` | **35 seconds**, GitHub-hosted |

## Nothing is unenforced by this, and that was checked first

Those nine invariants gate a merge through `ci.yml`'s `text-gates` job and its nine relay jobs, **every one of which is now a required context on `main`**.

They were not, until today. The ledger listed all nine; branch protection had none of them; and the gate that compares the two was itself not required, so it went red on a schedule with nothing forcing anyone to look. `38 live + 9 = 47 = PINNED`, and `xtask ci-spec live-parity` now reports *"ok: every invariant holds."*

What this deletes is a second, slower copy of the scans — not the scans.

The relays were already GitHub-hosted for exactly this reason (`ci.yml`: *"A relay on the pool costs a slot and undoes the consolidation"*). The scans hadn't followed them off the pool.

## What this is not

**Not the per-script scoping I measured earlier** (79 h → 27 h by giving each script its own scope). That optimised *where* the work ran instead of asking whether it belonged there at all.

**Not a judgement that the greps are weak.** They're backstops for what the stronger tiers structurally cannot see. `check-mediation.sh`'s own header says the primary enforcement is the type level — *"an un-preflighted effect is a compile error"* — and the grep catches the raw-primitive escape a type system cannot forbid. The ladder was already climbed; these are the rungs that can't be.

## Verification

`gate plan check` at the pinned `GATEHOUSE_REF`: 6 gates, admissibility proved, root kernel-checked. `xtask gatehouse-pin`: all three declarations agree. `just prepush`: 20/20.


brandon-coproduct:
Withdrawn. This would delete the gate that `ci/gatehouse-replacements.txt` points at — CI-RP-1 fires CRITICAL eleven times. The 79 gate-hours are real and still available, but as a deliberate reversal of the 2026-09-17 cutover, not as a tidy-up. See #2981.

Issue #2541, opened 2026-09-04
Host-side spend accounting via signed spend receipts verified at LedgerCore::release


brandon-coproduct:
Part of #2501 · Wave 3 · Size L · Blocked by #2526

## Problem
The budget ledger is host-side (`nucleus-node/src/pod_authority.rs:183,208`) while the 402 gate is guest-side in tool-proxy; a guest-side debit would be self-reported by the guest process.

## Fix
Tool-proxy emits **signed spend receipts** per granted dimension; `nucleus-node` verifies them at `LedgerCore::release` (`budget_ledger.rs:215`) when folding "actually spent" into `consumed`. No guest-side mutation of any budget.

## Acceptance criteria
- A forged or missing spend receipt cannot reduce or inflate the parent's consumed budget; conservation proofs from #2426 unchanged.

Issue #2540, opened 2026-09-04
Integer micro-USD λ pricing in nucleus-permission-market


brandon-coproduct:
Part of #2501 · Wave 3 · Size M · Blocked by #2526

## Problem
`compute_lambda` (`market.rs:198-208`) and grant totals are `f64` on a money path (`market.rs:130`, `main.rs:2585`).

## Fix
`compute_lambda_micro(utilization_bps: u32) -> u64`; `PermissionGrant` totals in micro-USD; delete the f64 API. Kani harness: overflow-free and monotone in utilization (pattern of `proofs/welfare_no_overflow.rs`).

## Acceptance criteria
- Parity test vs the old f64 within one micro-USD; `#![deny(clippy::float_arithmetic)]` on the crate.

Issue #2526, opened 2026-09-04
Remove the self-reported permission-bid path; bids derive only from the certificate budget


brandon-coproduct:
Part of #2501 · Wave 2 · Size M · Blocked by #2513

## Problem
`PermissionBid.value_estimate` (`crates/nucleus-permission-market/src/bid.rs:17-21`) is an arbitrary self-declared `f64` parsed from a header (`main.rs:~2627`); declaring a large value always wins at zero cost.

## Fix
Delete the header path; `PermissionBid` constructor becomes crate-private, built only by `cert_bridge::certificate_to_bid` (`cert_bridge.rs:46-94`) from the attested certificate budget; `trust_tier` sourced from verification, not the wire.

## Acceptance criteria
- `value_estimate` is never deserialized from a request; a structural test asserts no grant is computed without a certificate.

Issue #2986, opened 2026-09-21
scorecard bound census cannot see a path-qualified witness parameter


brandon-coproduct:
## What happened

While moving `certificate_to_bid(verified: &VerifiedPermissions)` from `crates/nucleus-tool-proxy/src/cert_bridge.rs` into `PermissionBid::from_verified(verified: &portcullis::VerifiedPermissions)` in `nucleus-permission-market` (#2526), `cargo xtask scorecard` reported:

```
bound's population fell to 170 from a floor of 171. Deleting an obligation raises the ratio without discharging anything ...
```

Nothing was deleted. The site had moved, and `crates/xtask/src/bound.rs` matches the witness type by its bare spelling, so `&portcullis::VerifiedPermissions` is not a site. Rewriting the parameter as `&VerifiedPermissions` with a `use` restored 171/170 with no code change.

## Why it matters

A witness-accepting site the census cannot see is one the ratchet does not protect (ADR 0007 I-1). Worse, the failure message steers the reader toward lowering `population_floor` with a dated note — which would have recorded a deletion that never happened.

## Fix

Match the last path segment (or resolve `use` aliases) when deciding whether a parameter type is a witness. Drive it red first (A-19): a fixture with a path-qualified `&portcullis::VerifiedPermissions` parameter must count.

Found during the #2526/#2540 rewrite.

PR #2987, opened 2026-09-21
A bid comes from a certificate, a price is an integer, and the node decides what was spent (#2526, #2540, #2541)

brandon-coproduct:
Closes #2526, closes #2540, closes #2541. Stacked on #2985 (→ #2977).

Three of epic #2501's children, as two commits.

## #2526 + #2540 — `e4f9278d4`

`PermissionBid` is sealed: private fields, no `Deserialize`, one constructor `from_verified(&VerifiedPermissions)`. The value is the principal's ceiling in micro-USD (truncated), the tier is `TrustTier::from_chain_depth`. The `x-nucleus-permission-bid` header path in the proxy is deleted; the no-certificate arm is `(None, None)`. The example plugin stops sending the header.

λ is a u128 fixed-point exponential — no float in the shipped build (`deny(float_arithmetic)` + the six other totality lints) — pinned to the f64 curve within **1 µ at every basis point** (the f64 oracle lives under `cfg(test)`). Prices round UP. Monotonicity is checked exhaustively over all 10 001 inputs under debug overflow checks; `proofs/lambda_monotone.rs` states it for Kani (crate pins `rust-version = "1.93"` like econ-kernels). The one f64 left is the 402 wire boundary, after an exact integer sum.

`PostedPriceClearing` / `RoundOutcome::PostedPrice` deleted from the exchange: it built a bid by struct literal, and it duplicated the proxy's own screen (G-1).

Ratchet: `tot` 3176 → 3294. (The `bound` census briefly read 171→170 on the move — it cannot see a path-qualified witness type; #2986.)

## #2541 — `64e3a7332`

`portcullis::spend_receipt::SpendReceipt`: pod, contiguous `seq`, micro-USD, clearing-receipt hash as basis, signed with the per-pod mediation key. The proxy ships one per won round as `SHIP_SPEND` (new vsock command; guest-init exports the port beside the key). The node verifies on arrival against its own `mediator-pubkey.hex` anchor, refuses by name, stores only what verified; `check` is the one decider and the fold re-runs it.

`release_child` folds `min(allocation, Σ verified)` **only** when receipts are exactly `1..=n`. `VerifiedSpend` is three-valued: no receipts and a gap both fold the full allocation, so suppressing a receipt costs more than shipping it. Test: one $4 admission that fails under the old rule and succeeds under the new, with no-receipts / gap / forged-key / over-claim controls.

The vsock walk ships a real signed body and writes the anchor iff a key is provisioned; the census gained the spend log as its own resource (faces 120→136, hollow still 10, non-idempotent gains `SHIP_SPEND`). Census test time roughly doubled (~2 min).

Trust granularity is the mediation key's (proxy, not workload) — stated, not solved.

## Verification

- `cargo test` full suites: nucleus-permission-market, nucleus-authority-exchange, nucleus-tool-proxy, nucleus-node, portcullis, nucleus-guest-init — green.
- clippy `-D warnings --all-targets` on all touched crates; `cargo xtask scorecard`, `cargo xtask econ-boundary` — green.
- First commit built standalone in a detached worktree.
- Kani: harness builds under `cargo kani --no-default-features`; the symbolic run had not terminated locally at PR time (u128 exp is heavy for CBMC) — the exhaustive plain test is the overflow/monotone evidence today.

https://claude.ai/code/session_01D5V6CuhLqCGCaLXwPzhtf8


brandon-coproduct:
## Closes #2988 too — and the composition is now proven on a live pod

Three commits added: the pod-peer tier, a probe diagnostic, and a startup warning.

### What was broken

The exchange and #2541's spend receipts were reachable on **disjoint tiers**: a contested round needed ≥2 distinct mTLS callers; spend receipts needed a Firecracker guest, where the proxy serves vsock, the vsock listener admits *only the host*, and every in-guest caller is one unbound identity. An in-guest workload could not reach its own proxy at all.

### The fix

The peer-verified Unix socket may now bind **beside** vsock. The exclusivity rule existed so the audit record named the transport; it still does, per connection: each accepted Unix connection carries the kernel-reported `(uid, pid)` as `ConnectInfo<PodPeer>`, and an in-namespace peer is its own tier (`AuthTier::PodPeer` / `AuthMethod::PodPeer`), ranked under the certificate and approval tiers, above `HostVsock`. The host through that socket (pid 0) stays `HostVsock`. `delegation_authority` keeps the tier **Unbound** — a uid is not a SPIFFE leaf — so a certificate header on it is refused exactly as before.

`authority_round::Bidder` decides who bids from facts a request cannot claim: a verified certificate (own ceiling, own leaf), or a pod peer, which bids under the **pod's** sealed `VerifiedPermissions` as `<pod leaf>/peer/uid-N/pid-M`. Two processes are two bidders; one process is one. A bidder may declare a value *below* its ceiling (`x-nucleus-bid-micro-usd`); above is refused, never clamped.

### Live run (nucleus-kvm, real Firecracker pod, 3 workload processes)

```
bid=1000000 pid=104 → outbid for the network_egress slot in round NetworkEgress-1
bid=2000000 pid=105 → outbid
bid=3000000 pid=106 → won → then 403: web_fetch "available Never"
SUMMARY children=3 won=1 outbid=2 denied=0 contested=true
```

- Second-price clearing: the winner was charged **2 000 000 µUSD**, the second-highest bid, not its own.
- **Economics did not widen authority**: the winner still hit the capability lattice.
- The signed `SpendReceipt` reached the host and verified against the node's own `mediator-pubkey.hex`:
  `{"seq":1,"amount_micro":2000000,"basis":"authority-round:4aae534a…"}`
- At release the node logged
  `verified spend receipts decide the released budget receipts=1 spent_usd=2.000000 allocation_usd=5.00`
  — folding $2.00 instead of the full $5.00 allocation. **Control in the same log**: earlier pods with no receipts got `no verified spend receipts; the full allocation is consumed`.

`contested=true` is the non-vacuity gate — the probe exits 1 if nobody was outbid.

### One defect the run found, fixed here

With the socket under `/run/nucleus` — mode **700 root**, because it holds the pod's SVID — every workload child was refused at connect with EACCES and the proxy said nothing: `--peer-uids` was an allowlist admitting nobody. `bind_unix` now walks the ancestors and names the first directory that excludes other uids. A warning, not a refusal: the operator may have chowned it.

https://claude.ai/code/session_01D5V6CuhLqCGCaLXwPzhtf8

brandon-coproduct:
## Correction: the λ Kani harness is gone

My earlier comment claimed a Kani harness for `compute_lambda_micro`. It is removed, and the claim was already weak when I made it — I noted then that the symbolic run "had not terminated locally".

The proof-count gate made the issue concrete:

```
Error: uncovered harness: lambda_monotone.rs::lambda_is_not_zero_everywhere
       uncovered harness: lambda_monotone.rs::lambda_monotone_and_bounded
```

**A harness no workflow schedules is a proof nobody checks**, and I could not honestly schedule these: CBMC spent fifty minutes on the `u128` multiplications without terminating. Scheduling it would have bought a timeout and the word "Kani" in a table.

It was also redundant. `compute_lambda_micro` is straight-line over a finite domain — `bps` is capped at 10 000 — so the existing test walks **all 10 001 inputs** under debug overflow checks and decides monotonicity and overflow-freedom outright, for the code that ships. That is exhaustive; a bounded model checker would re-prove the same finite facts more slowly. The parity test still pins the curve to within one micro-unit of the `f64` curve it replaced at every basis point.

Removed with it: the crate's 1.93 MSRV pin (Kani's toolchain was its only justification) and the `cfg(kani)` lint exemption. The census returns to 120 in all five files that record it, `.kani-minimum-proofs`, and the divergence inventory.

## Triage summary

Sixteen red checks, all mine, five distinct causes — none of them in the pod-peer mechanism itself:

| Cause | Fix |
|---|---|
| `xtask econ-boundary` had no perturbation probe (my own #2514 gate) | registered in `check-gates-can-fail.sh` |
| The Kani census is written in five doc files, not just the pin | see above — reverted to 120 |
| `spend_shipper` is a second raw vsock client | mediation allowlist, pin 1 → 2 |
| Four cast sites over the ratchet | removed, not allowed: 343 → 338 |
| `permissive_verify` 13 → 14 | `SpendReceipt::verify_strict`, so the guarantee reaches call sites |
| `main.rs` two lines over its ceiling | connect-info read moved to `host_socket::pod_peer_of` |

Two of those I found only because the gate-of-gates reported them as "already red BEFORE any perturbation" — including one missed call site in a test, which `cargo build -p` does not compile.

https://claude.ai/code/session_01D5V6CuhLqCGCaLXwPzhtf8

